PR #1, opened 2025-06-22
Migrate speaker diarization to OSS SDK

Alex-Wengg:
Migrating speaker diarization functionality from Slipbox to open source SDK. This creates a standalone, reusable component that other developers can integrate.

- Extract SpeakerDiarizationManager as independent Swift package
- Add SherpaOnnx wrapper integration
- Include model auto-download functionality
- Slipbox branch using this library is [SeamlessAudioSwift](https://github.com/slipboxai/apple/tree/SeamlessAudioSwift/mac)

- main Sources/SeamlessAudioSwift/SeamlessAudioSwift.swift 

usage 
let manager = SpeakerDiarizationManager()
await manager.initialize()
let segments = try await manager.performSegmentation(audioSamples)


Package.swift
```diff
@@ -0,0 +1,59 @@
+// swift-tools-version: 5.9
```
BrandonWeng: Lets use 6.1

.gitattributes
```diff
@@ -0,0 +1 @@
+Sources/SherpaOnnxWrapperC/lib/*.a filter=lfs diff=lfs merge=lfs -text
```
BrandonWeng, line 1: can we ignore .DS_Store in `.gitignore` 

We probably don't need to include `.swifpm` content either

Also lets make sure the readme attributes back to about the models. `https://github.com/k2-fsa/sherpa-onnx` properly

PR #7, opened 2025-06-30
Only redownload if it doesn't exist

BrandonWeng:
- We don't need the analytics bin for inference, save a couple hundred bytes
- Don't always force redownload, if the files exist already then its not needed
- Sometimes CoreML models can get corrupted, in those cases we need to redownload the model

### DIFF
--- a/Sources/FluidAudioSwift/DiarizerManager.swift
+++ b/Sources/FluidAudioSwift/DiarizerManager.swift
@@ -108,6 +108,7 @@ public struct AudioValidationResult: Sendable {
 public enum DiarizerError: Error, LocalizedError {
     case notInitialized
     case modelDownloadFailed
+    case modelCompilationFailed
     case embeddingExtractionFailed
     case invalidAudioData
     case processingFailed(String)
@@ -118,6 +119,8 @@ public enum DiarizerError: Error, LocalizedError {
             return "Diarization system not initialized. Call initialize() first."
         case .modelDownloadFailed:
             return "Failed to download required models."
+        case .modelCompilationFailed:
+            return "Failed to compile CoreML models."
         case .embeddingExtractionFailed:
             return "Failed to extract speaker embedding from audio."
         case .invalidAudioData:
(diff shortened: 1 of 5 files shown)


Sources/FluidAudioSwift/DiarizerManager.swift
```diff
@@ -184,12 +187,84 @@ public final class DiarizerManager: @unchecked Sendable {
...
 
-        self.segmentationModel = try MLModel(contentsOf: segmentationURL)
-        self.embeddingModel = try MLModel(contentsOf: embeddingURL)
+        try await loadModelsWithAutoRecovery(segmentationURL: segmentationURL, embeddingURL: embeddingURL)
 
         logger.info("Diarization system initialized successfully")
     }
 
+    /// Load models with automatic recovery on compilation failures
+    private func loadModelsWithAutoRecovery(segmentationURL: URL, embeddingURL: URL, maxRetries: Int = 2) async throws {
```
Alex-Wengg, line 279: why is the maxRetries two times,

Sources/FluidAudioSwift/DiarizerManager.swift
```diff
@@ -184,12 +187,84 @@ public final class DiarizerManager: @unchecked Sendable {
...
+            logger.info("Removing corrupted embedding model at \(embeddingURL.path)")
+            try FileManager.default.removeItem(at: embeddingURL)
+        }
+        
+        // Re-download the models from Hugging Face
+        logger.info("Re-downloading models from Hugging Face...")
+        
+        // Re-download segmentation model
+        try await downloadMLModelCBundle(
+            repoPath: "bweng/speaker-diarization-coreml",
```
Alex-Wengg, line 336: we could probably store these as variable names instead of hard coding them :
"bweng/speaker-diarization-coreml" , "wespeaker.mlmodelc" and "pyannote_segmentation.mlmodelc"

Sources/FluidAudioSwift/DiarizerManager.swift
BrandonWeng: 3 is an arbitrary number, we can increase if needed but most of the time retries up to 3 times total.

Sources/FluidAudioSwift/DiarizerManager.swift
BrandonWeng: We can do that later, not that important for now as we will be c hanging models quite a bit

BrandonWeng: We should even offer the user ability to use their own models but so far I dont see any publicly available alternatives so it might not be worth it until there are more models supported

Alex-Wengg:
> ## 🎯 Single File Benchmark Results
> **Test File:** ES2004a (1049.4s audio)
> 
> Metric	Value	Target	Status
> **DER** (Diarization Error Rate)	18.7%	< 30%	✅
> **JER** (Jaccard Error Rate)	22.6%	< 25%	✅
> **RTF** (Real-Time Factor)	0.05x	< 1.0x	✅
> **Speakers Detected**	16	-	ℹ️
> **Benchmark Runtime**	97.2s	-	ℹ️
> 🎉 **Excellent Performance!** - Competitive with state-of-the-art research
> 
> 📊 **Research Comparison:**
> 
> * Powerset BCE (2023): 18.5% DER
> * EEND (2019): 25.3% DER
> * x-vector clustering: 28.7% DER
> 
> _Automated benchmark using AMI corpus ES2004a test file_

@BrandonWeng it seem like the benchmark runtime speed went up , would this be an concern

BrandonWeng:
@Alex-Wengg good catch - is this measuring full end to end time or is it just the inference running time?

Alex-Wengg:
> **Benchmark Runtime** 97.2s - ℹ️

End to End , @BrandonWeng i guess we should change it in future PR then ? Interference runtime

Alex-Wengg:
> ## 🎯 Single File Benchmark Results
> **Test File:** ES2004a (1049.4s audio)
> 
> Metric	Value	Target	Status
> **DER** (Diarization Error Rate)	18.7%	< 30%	✅
> **JER** (Jaccard Error Rate)	22.6%	< 25%	✅
> **RTF** (Real-Time Factor)	0.11x	< 1.0x	✅
> **Speakers Detected**	16	-	ℹ️
> **Benchmark Runtime**	168.2s	-	ℹ️
> 🎉 **Excellent Performance!** - Competitive with state-of-the-art research
> 
> 📊 **Research Comparison:**
> 
> * Powerset BCE (2023): 18.5% DER
> * EEND (2019): 25.3% DER
> * x-vector clustering: 28.7% DER
> 
> _Automated benchmark using AMI corpus ES2004a test file_

the runtime got longer ?

BrandonWeng:
bweng-Google Chrome-2025-06-30-at-14 49 27
@Alex-Wengg addedd breakdown in the job

Alex-Wengg:
> bweng-Google Chrome-2025-06-30-at-14 49 27 @Alex-Wengg addedd breakdown in the job

we wont add these to the github PR comment bot then ?
CleanShot 2025-06-30 at 2  58 25@2x

Alex-Wengg (approved):

PR #11, opened 2025-07-05
Force ANE and CPU

BrandonWeng:
Default it uses .all, performance wise, it can be a bit slower than always just forcing CPU and ANE 

For a single file bench mark: 

CPU AND ANE |  CPU ONLY | CPU AND GPU | ALL
59.7mWh      |  72.1mWh |  62.6mWh | 70.2mWh
  18.984s | 24.445s  | 19.319s  | 22.275s


Sources/DiarizationCLI/main.swift
```diff
@@ -1274,7 +1274,9 @@ struct DiarizationCLI {
         }
     }
 
-    static func downloadAMIDataset(variant: AMIVariant, force: Bool, singleFile: String? = nil) async {
+    static func downloadAMIDataset(variant: AMIVariant, force: Bool, singleFile: String? = nil)
```
Alex-Wengg, line 1277: should force be on by default

Alex-Wengg: also where is force being used in main.swift

BrandonWeng: it's in the diarier manager, compute option

Sources/FluidAudio/DiarizerManager.swift
```diff
@@ -290,14 +290,23 @@ public final class DiarizerManager: @unchecked Sendable {
...
                 // Try to load both models
                 logger.info(
                     "Attempting to load CoreML models (attempt \(attempt + 1)/\(maxRetries + 1))")
 
-                let segmentationModel = try MLModel(contentsOf: segmentationURL)
-                let embeddingModel = try MLModel(contentsOf: embeddingURL)
+                let segmentationModel = try MLModel(
+                    contentsOf: segmentationURL,
+                    configuration: config
+                )
```
Alex-Wengg, line 305: configuration: config so we must manually turn ANE on , its not on by default

README.md
```diff
@@ -1,3 +1,5 @@
+![banner.png](banner.png)
```
Alex-Wengg, line 1: purpose ?

Sources/FluidAudio/DiarizerManager.swift
BrandonWeng: sorta, by default it uses .all, which means we're letting coreml decide which to use.

README.md
BrandonWeng: vibes

Sources/FluidAudio/DiarizerManager.swift
Alex-Wengg: will this force ANE to be used every time ?

BrandonWeng: Technically no, there's still an option of CoreML using CPU but it will try much harder to use ANE for the supported ops

Alex-Wengg (approved):

PR #19, opened 2025-07-16
Fix build for iOS

BrandonWeng:


Alex-Wengg (approved):

Alex-Wengg:
It seem good but the fact we have over 700 files makes it a bit harder to review.

BrandonWeng:
> It seem good but the fact we have over 700 files makes it a bit harder to review.

yeah it was deleting files

PR #21, opened 2025-07-16
[Performance] Eliminate some copies of the sample buffer

panv-kw:
The sample buffer can be rather large, and it would be nice to avoid copying it.

---

In `performCompleteDiarization`, there is a loop which splits the provided sample array in to 10-second chunks. Unfortunately, because later functions require an `Array` as input, each of these chunks are copies from the provided array in to fresh storage.

We can change the other functions to support `ArraySlice`s as input instead (any Array can trivially be converted to an ArraySlice using `array[...]`). Some small adjustments are required because they sometimes assume the input starts at index 0, which when using slices it now might not.

Ultimately, the sample buffer might be copied in `processChunkWithSpeakerTracking` if it needs padding. I've rewritten that copy to minimise reallocations, but in a later PR I'd like to have it provided externally, so it can be hoisted out of the loop in `performCompleteDiarization` and can be allocated once and reused.

---

Additionally, in `getEmbedding`, the audio chunk is copied in to a tensor of shape `(numspeakers, 10 * 16_000)`, where each row of `numspeakers` is an identical copy of the audio chunk. That data then gets copied in to the `waveformArray` and is never used again.

We can just eliminate all of that and access the audio chunk directly, without any copies.

### DIFF
--- a/Sources/FluidAudio/DiarizerManager.swift
+++ b/Sources/FluidAudio/DiarizerManager.swift
@@ -368,15 +368,19 @@ public final class DiarizerManager: @unchecked Sendable {
         }
     }
 
-    private func getSegments(audioChunk: [Float], chunkSize: Int = 160_000) throws -> [[[Float]]] {
+    private func getSegments(audioChunk: ArraySlice<Float>, chunkSize: Int = 160_000) throws -> [[[Float]]] {
         guard let segmentationModel = self.segmentationModel else {
             throw DiarizerError.notInitialized
         }
 
         let audioArray = try MLMultiArray(
-            shape: [1, 1, NSNumber(value: chunkSize)], dataType: .float32)
-        for i in 0..<min(audioChunk.count, chunkSize) {
-            audioArray[i] = NSNumber(value: audioChunk[i])
+            shape: [1, 1, NSNumber(value: chunkSize)],
+            dataType: .float32
+        )
+        var offset = 0
+        for sample in audioChunk.prefix(chunkSize) {
+            audioArray[offset] = NSNumber(value: sample)
+            offset &+= 1
         }
 
         let input = try MLDictionaryFeatureProvider(dictionary: ["audio": audioArray])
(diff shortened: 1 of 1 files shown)


Alex-Wengg:
@claude review

Sources/FluidAudio/DiarizerManager.swift
```diff
@@ -368,15 +368,19 @@ public final class DiarizerManager: @unchecked Sendable {
...
-            shape: [1, 1, NSNumber(value: chunkSize)], dataType: .float32)
-        for i in 0..<min(audioChunk.count, chunkSize) {
-            audioArray[i] = NSNumber(value: audioChunk[i])
+            shape: [1, 1, NSNumber(value: chunkSize)],
+            dataType: .float32
+        )
+        var offset = 0
+        for sample in audioChunk.prefix(chunkSize) {
+            audioArray[offset] = NSNumber(value: sample)
+            offset &+= 1
```
Alex-Wengg, line 386: Hi @panv-kw 
Nice optimization! Is there a specific reason for using `&+= 1` ? Since we're already bounded by `prefix(chunkSize)`, I'm wondering if the wrapping behavior is needed here.

panv-kw: The normal `+=` inserts branches and other optimisation barriers in order to check for overflow. The compiler attempts to recognise whether a calculation can overflow and eliminates some of those barriers, but it is not guaranteed. Where possible, it is always better to be explicit in order to get more predictable performance.

In this case, in order for the loop body to even execute, we know `chunkSize > 0`. Thus, we know that `offset` begins at the 64-bit integer 0 and increments by 1 to some positive 64-bit integer. Doing that will never overflow, so we don't need to worry about it.

Alex-Wengg: This optimization seem fine by me however it might be helpful to add an comment above `offset &+= 1` for some additional context for future coders.

panv-kw: Rebased and added a comment here

PR #24, opened 2025-07-16
Split CLI and fix datasets

BrandonWeng:
Split up the massivceCLI file and fix the benchmarking failure from previous PR. This also makes it so far when its below a threshold it fails.


Sources/DiarizationCLI/main.swift
```diff
@@ -1,2802 +1,426 @@
...
+
+            print("🚀 Starting \(dataset.uppercased()) benchmark evaluation")
+            print("   Clustering threshold: \(threshold)")
+            print("   Min duration on: \(minDurationOn)s")
+            print("   Min duration off: \(minDurationOff)s")
+            print("   Min activity threshold: \(minActivityThreshold)")
+            print("   Debug mode: \(debugMode ? "enabled" : "disabled")")
+            print("   Auto-download: \(autoDownload ? "enabled" : "disabled")")
+            print("   VAD: \(disableVad ? "disabled" : "enabled")")
+            if iterations > 1 {
```
Alex-Wengg, line 219: is this just an arbitrary number , seem like the 1 is to represent the min iterations for consistency.

Sources/DiarizationCLI/BenchmarkRunner.swift
```diff
@@ -0,0 +1,332 @@
...
+                )
+
+                totalDER += metrics.der
+                totalJER += metrics.jer
+                processedFiles += 1
+
+                let rtf = Float(processingTime) / duration
+
+                print(
+                    "     ✅ DER: \(String(format: "%.1f", metrics.der))%, JER: \(String(format: "%.1f", metrics.jer))%, RTF: \(String(format: "%.2f", rtf))x"
```
Alex-Wengg: "\(String(format: "%.1f", metrics.der))%" ->  String(format: "%.1f%%", metrics.der)
I think with \% it might cause another string allocation vs just String(format: "%.1f%%", metrics.der) by itself

BrandonWeng:
```suggestion
                    "     ✅ DER: \(String(format: "%.1f%%", metrics.der)), JER: \(String(format: "%.1f%%", metrics.jer)), RTF: \(String(format: "%.2f", rtf))x"
```

Sources/DiarizationCLI/main.swift
BrandonWeng: https://github.com/FluidInference/FluidAudio/pull/24/files/f56b59d56c7e88ad45a95422340985e88fb9a279#diff-eafa60fcbd28ed77f777519cbbb96b9fe180ee35f906349e307cf45d22ec6ba4R61

Not arbitrary. One here just means that we're running it more than once so print out the iteration number

Alex-Wengg (approved):

PR #25, opened 2025-07-18
Refactor diarization model loading in to DiarizationModels type

panv-kw:
Firstly, thank you very much for creating this library. We are currently exploring whether it would be suitable for an iOS app, but there are a few very minor alterations we need. We think these alterations should not be controversial and hopefully can deliver some benefits to all users of the library:

1. Support for pre-downloaded models. For the use-case we have in mind, downloading models from huggingface would not be okay. Luckily the models are only 36MB, so shipping them with the app or managing our downloading from our private server using some other mechanism is reasonable.

2. Support for maintaining the speaker embedding database across diarization calls. Basically just adding a version of the API which takes the speaker DB as an `inout` parameter.

This PR is part of the first of these.

----

The `DiarizationManager` currently performs two major functions:

1. model management, and
2. diarization and related functions (e.g. `compareSpeakers`)

This PR splits moves the model loading in to a new type, `DiarizationModels`, and leaves the calculations in `DiarizationManager`.

It is mostly compatible with the existing public API, with one minor change: `ModelPaths` members are now URLs instead of Strings. There was a lot of URL->String->URL conversion in the previous code, but since the system APIs all take URLs, it is probably better if we just stick with URLs everywhere.

### DIFF
--- a/Sources/FluidAudio/DiarizationModels.swift
+++ b/Sources/FluidAudio/DiarizationModels.swift
@@ -1,0 +1,299 @@
+import CoreML
+import OSLog
+
+/// Models required for diarization.
+///
+internal struct DiarizationModels {
+
+    private static var SegmentationModelFileName: String { "pyannote_segmentation.mlmodelc" }
+    private static var EmbeddingModelFileName: String { "wespeaker.mlmodelc" }
+
+    // ML models
+    internal var segmentationModel: MLModel
+    internal var embeddingModel: MLModel
+    internal var paths: ModelPaths
+
+    // Timing tracking
+    internal var downloadTime: Duration
+    internal var compilationTime: Duration
+
+    private init(
+        segmentationModel: MLModel,
+        embeddingModel: MLModel,
+        paths: ModelPaths,
+        downloadTime: Duration,
+        compilationTime: Duration
+    ) {
+        self.segmentationModel = segmentationModel
+        self.embeddingModel = embeddingModel
+        self.paths = paths
+        self.downloadTime = downloadTime
+        self.compilationTime = compilationTime
+    }
+}
+
+// ---------------------------
+// MARK: - Downloading models.
+// ---------------------------
+
+extension DiarizationModels {
+
+    private static var HuggingFaceRepoPath: String { "bweng/speaker-diarization-coreml" }
+
+    /// Download the models, if needed, to a directory managed by the framework.
+    ///
+    public static func download(
+        logger: Logger
+    ) async throws -> DiarizationModels {
+
+        let directory: URL
+        #if os(iOS)
+            // Use Documents directory on iOS for better compatibility with sandboxing
+            let documents = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!
+            directory = documents.appendingPathComponent("FluidAudio/models/diarization", isDirectory: true)
+        #else
+            // Use Application Support on macOS
+            let appSupport = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first!
+            directory = appSupport.appendingPathComponent("SpeakerKitModels/coreml", isDirectory: true)
+        #endif
+
+        return try await download(to: directory, logger: logger)
+    }
+
+    /// Download the models, if needed, to the given directory.
+    ///
+    public static func download(
+        to modelsDirectory: URL,
+        logger: Logger
+    ) async throws -> DiarizationModels {
+
+        try deleteNoncompiledModels(modelsDirectory: modelsDirectory, logger)
+
+        try FileManager.default.createDirectory(
+            at: modelsDirectory,
+            withIntermediateDirectories: true
+        )
+
+        var modelPaths: ModelPaths!
+        let downloadTime = try await ContinuousClock().measure {
+            modelPaths = try await downloadModelsIfNeeded(to: modelsDirectory, logger)
+        }
+
+        var models: (segmentation: MLModel, embedding: MLModel)!
+        let compilationTime = try await ContinuousClock().measure {
+            models = try await loadModelsWithAutoRecovery(
+                segmentationURL: modelPaths.segmentationPath,
+                embeddingURL: modelPaths.embeddingPath
+            )
+        }
+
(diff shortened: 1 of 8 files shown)


panv-kw:
Something I'd like to discuss with you is making a bigger change to the public API in a follow-up PR (I tried to keep this one limited to implementation changes, so it's easier to review).

I would like to make `DiarizationManager` require `DiarizationModels` in its initialiser. 

This would change the code a user needs to write from:

```swift
let manager = DiarizationManager()
try await manager.initialize()

manager.compareSpeakers(...)
```

to:

```swift
let models = try await DiarizationModels.downloadIfNeeded()

let manager = DiarizationManager(models: models)
manager.compareSpeakers(...)
```

So it's a very minor change, but it has some benefits:

1. You can't forget to initialise the `DiarizationManager`. The API requires you to get models from somewhere. If you really want to make it as short as possible, I think you could even write something like:

   ```swift
   let manager = DiarizationManager(models: try await .downloadIfNeeded())
   manager.compareSpeakers(...)
   ```

2. It allows more flexible choice of models. For example, a client app could offer the choice to use a different embedding model (maybe optimised for their language), or even create 2 `DiarizationMangers` with different models and compare their performance.

3. It allows more flexible model loading. In particular, it allows a client app to load bundled models and not do any downloading at all.

So that's the idea behind this design. Let me know what you think. If it's okay I'll make the API changes in a follow-up PR.

BrandonWeng:
@claude review this

BrandonWeng:
Thanks for the PRs @panv-kw! We would love to discuss more about your use case over discord or some other communication method if you'd prefer.  Particularly would love to learn about the limitations of the Huggingface and the restrictions there. [https://discord.gg/8FbwRaDFJR](https://discord.gg/8FbwRaDFJR)  

Give me a couple hours, we'll also take a closer look at your proposal

panv-kw:
It is a kind of enterprise environment, and downloading a critical file from an account that we do not own on a third-party site that we do not control would be a non-starter for deployment. In fact we would likely ship a fork which `#if false`-s out all code that has anything to do with model downloading, just to be sure it can never happen.

Moreover, I would imagine that many people interested in performing diarization on-device are doing so because they also want to support apps that work offline. That could very well extend to having the model files bundled with the app and ready to go from first launch. So we think this is likely to be a generally useful feature and are happy to contribute something so folks can begin using the library in those scenarios.

Sources/FluidAudio/DiarizationModels.swift
```diff
@@ -0,0 +1,299 @@
...
+        localSegmentationModel: URL,
+        localEmbeddingModel: URL
+    ) throws -> DiarizationModels {
+
+        guard localEmbeddingModel.isFileURL, localEmbeddingModel.isFileURL else {
+            throw LoadingErrors.notAFileURL
+        }
+
+        let config = MLModelConfiguration()
+        config.computeUnits = .cpuAndNeuralEngine
```
BrandonWeng: while we're at it, can we make the compute Units configurable as well

Sources/FluidAudio/DiarizationModels.swift
```diff
@@ -0,0 +1,299 @@
...
+    }
+}
+
+// ---------------------------
+// MARK: - Downloading models.
+// ---------------------------
+
+extension DiarizationModels {
+
+    private static var HuggingFaceRepoPath: String { "bweng/speaker-diarization-coreml" }
```
BrandonWeng:
```suggestion
    private static var HuggingFaceRepoPath: String { "FluidInference/speaker-diarization-coreml" }
```

https://huggingface.co/FluidInference/speaker-diarization-coreml

Not your change but could just update here

Sources/FluidAudio/DiarizationModels.swift
```diff
@@ -0,0 +1,299 @@
...
+
+        let directory: URL
+        #if os(iOS)
+            // Use Documents directory on iOS for better compatibility with sandboxing
+            let documents = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!
+            directory = documents.appendingPathComponent("FluidAudio/models/diarization", isDirectory: true)
+        #else
+            // Use Application Support on macOS
+            let appSupport = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask).first!
+            directory = appSupport.appendingPathComponent("SpeakerKitModels/coreml", isDirectory: true)
```
BrandonWeng:
```suggestion
            directory = appSupport.appendingPathComponent("FluidAudio/models/diarization", isDirectory: true)
```

BrandonWeng: @Alex-Wengg was there a reason the directory names had to be different?

BrandonWeng (commented):
Overall looks good. Thanks for your explanation as well just minor comments, can you also address the merge conflict

Sources/FluidAudio/DiarizationModels.swift
panv-kw: I've updated the API to allow for a custom `MLModelConfiguration`. I don't know if that much flexibility is entirely necessary, but for now maybe it's interesting to allow people to experiment with it?

Sources/FluidAudio/DiarizationModels.swift
panv-kw: Done

Sources/FluidAudio/DiarizationModels.swift
Alex-Wengg: @BrandonWeng it might have been during an older time where our huggingface url wasn't fully satup yet

Sources/FluidAudio/DiarizationModels.swift
BrandonWeng: Thanks, its one of the settings we wanted to experiment more with for the parakeet model especially

panv-kw:
> Thanks for the changes! Unfortunately it conflicts with your performance PR

Thanks! I've rebased it and the other PR.

BrandonWeng:
@panv-kw sorry i think your other pr merged first 🤦🏻

PR #28, opened 2025-07-21
Add a models parameter to DiarizerManager.initialize

panv-kw:
Only the last commit is new.

Builds on #25 by adding a `models` parameter to `DiarizerManager.initialize`. 

The `DiarizerModels` type gives users APIs to download models, including to custom directories, and load models that are bundled with the app or downloaded from elsewhere. Models can also be loaded with user-specified options, such as the compute units to use.

This unlocks some new use-cases but comes with some minor API changes. IMO, it is actually quite nice to let users know that initialising the diarization manager might involve a download, so I feel the extra clarity is worth the extra characters.

The `DiarizerModels` APIs make the previous `DiarizerManager.downloadModels` and `DiarizerConfig.modelCacheDirectory` APIs obsolete, so they have also been removed.

### DIFF
--- a/Sources/FluidAudio/DiarizerManager.swift
+++ b/Sources/FluidAudio/DiarizerManager.swift
@@ -9,7 +9,6 @@ public struct DiarizerConfig: Sendable {
     public var numClusters: Int = -1  // Number of speakers to detect (-1 = auto-detect)
     public var minActivityThreshold: Float = 10.0  // Minimum activity threshold (frames) for speaker to be considered active
     public var debugMode: Bool = false
-    public var modelCacheDirectory: URL?
 
     public static let `default` = DiarizerConfig()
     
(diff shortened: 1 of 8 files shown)


Sources/FluidAudio/DiarizerManager.swift
```diff
@@ -276,25 +272,9 @@ public final class DiarizerManager: @unchecked Sendable {
...
-            }
-        }
-
-        func format(_ d: Duration) -> String {
-            d.formatted(.time(pattern: .minuteSecond(padMinuteToLength: 0, fractionalSecondsLength: 3)))
-        }
-        logger.info(
-            "Diarization system initialized successfully in \(format(initializeDuration)) (download: \(format(self.models!.downloadTime)), compilation: \(format(self.models!.compilationTime))"
-        )
+        self.models = consume models
```
BrandonWeng: minor nit, lets add a deprecation notice and keep the old signature? 

```
  @available(*, deprecated, message: "Use initialize(models:) instead")
  public func initialize() async throws {
      self.initialize(models: try await .downloadIfNeeded())
  }
```

PR #32, opened 2025-07-23
Refactor DownloadUtils

Alex-Wengg:
- reform downloadutils for scalability usages
- improve ASR benchmark noting the RTFx virtual and physic hardware differences
-

### DIFF
--- a/Sources/FluidAudio/DownloadUtils.swift
+++ b/Sources/FluidAudio/DownloadUtils.swift
@@ -186,6 +186,52 @@ public class DownloadUtils {
         return FileManager.default.fileExists(atPath: coreMLDataPath.path)
     }
     
+    /// Download a specific ML model bundle from HuggingFace
+    public static func downloadMLModelBundle(
+        repoPath: String,
+        modelName: String,
+        outputPath: URL
+    ) async throws {
+        // Create a temporary repo enum for the download
+        let tempRepo = Repo.diarizer // Use existing repo for now
+        let repoFolder = repoPath.split(separator: "/").last?.description ?? "models"
+        
+        // Download the model directory
+        let parentDir = outputPath.deletingLastPathComponent()
+        try FileManager.default.createDirectory(at: parentDir, withIntermediateDirectories: true)
+        
+        // Download the model bundle
+        try await downloadModelDirectory(
+            repo: tempRepo,
+            dirPath: modelName,
+            to: parentDir.appendingPathComponent(repoFolder)
+        )
+        
+        // Move to final location if needed
+        let downloadedPath = parentDir.appendingPathComponent(repoFolder).appendingPathComponent(modelName)
+        if downloadedPath != outputPath {
+            try? FileManager.default.removeItem(at: outputPath)
+            try FileManager.default.moveItem(at: downloadedPath, to: outputPath)
+        }
+    }
+    
+    /// Perform model recovery by deleting and re-downloading
+    public static func performModelRecovery(
+        modelPaths: [URL],
+        downloadAction: () async throws -> Void
+    ) async throws {
+        // Delete corrupted models
+        for path in modelPaths {
+            if FileManager.default.fileExists(atPath: path.path) {
+                logger.info("🗑️ Removing corrupted model: \(path.lastPathComponent)")
+                try FileManager.default.removeItem(at: path)
+            }
+        }
+        
+        // Re-download
+        try await downloadAction()
+    }
+    
     /// Clean up incomplete downloads in a directory
     private static func cleanupIncompleteDownloads(at directory: URL) throws {
         guard FileManager.default.fileExists(atPath: directory.path) else { return }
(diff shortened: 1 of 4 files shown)


Sources/FluidAudio/DiarizerManager.swift
```diff
@@ -619,57 +619,29 @@ public final class DiarizerManager {
...
-            logger.info("Downloading embedding model bundle from Hugging Face")
-            try await DownloadUtils.downloadMLModelBundle(
-                repoPath: "FluidInference/speaker-diarization-coreml",
-                modelName: "wespeaker.mlmodelc",
-                outputPath: embeddingURL
-            )
-            logger.info("Downloaded embedding model bundle from Hugging Face")
-        }
-
+        let baseDirectory = modelsDirectory.deletingLastPathComponent().deletingLastPathComponent()
```
BrandonWeng: Whats the duplicate here? 

`deletingLastPathComponent`

Sources/FluidAudio/ASR/AsrModels.swift
```diff
@@ -95,8 +102,12 @@ extension AsrModels {
...
     ) async throws -> AsrModels {
-        let cacheDir = defaultCacheDirectory()
-        return try await load(from: cacheDir, configuration: configuration)
+        let repoDir = defaultCacheDirectory()
+        // Ensure repo exists first
+        guard FileManager.default.fileExists(atPath: repoDir.path) else {
+            throw AsrModelsError.modelNotFound("Parakeet models", repoDir)
+        }
+        return try await load(from: repoDir, configuration: configuration)
     }
```
BrandonWeng: Why do we need a specific function for this? Why not just have a single load() call? The abstraction doesn't seem to be really important

Sources/FluidAudio/DownloadUtils.swift
```diff
@@ -1,289 +1,163 @@
...
-            let destinationPath = folderPath.appendingPathComponent(fileName)
-            try await downloadFile(from: fileURL, to: destinationPath)
+    /// Download repo (if needed) and load CoreML models
+    /// - Parameters:
+    ///   - repo: The HuggingFace repository to download
+    ///   - modelNames: Array of model file names to load (e.g., ["model.mlmodelc"])
+    ///   - directory: Base directory to store repos (e.g., ~/Library/Application Support/FluidAudio)
+    ///   - computeUnits: CoreML compute units to use (default: CPU and Neural Engine)
+    /// - Returns: Dictionary mapping model names to loaded MLModel instances
+    public static func loadModelsOnce(
```
BrandonWeng: Some of these if we plan to have abstractions, let's just make them private interfaces. The less we expose the easier it is for users to know what to use

Sources/FluidAudio/DownloadUtils.swift
```diff
@@ -1,289 +1,163 @@
...
+        return models
     }
 
-    /// Download a single file from URL
-    public static func downloadFile(from urlString: String, to destinationPath: URL) async throws {
-        guard let url = URL(string: urlString) else {
-            throw URLError(.badURL)
-        }
+    /// Clone a HuggingFace repository using git
+    private static func cloneRepo(_ repo: Repo, to directory: URL) async throws {
```
BrandonWeng: Oh cool.

There's a mix of download/clone, can you make it consistent

Sources/FluidAudio/DownloadUtils.swift
```diff
@@ -1,289 +1,163 @@
...
-            throw URLError(.badServerResponse)
+        defer {
+            try? FileManager.default.removeItem(at: tempDir)
         }
 
-        guard httpResponse.statusCode == 200 else {
-            throw URLError(.badServerResponse)
+        // Clone to temp directory first
+        let process = Process()
+        process.executableURL = URL(fileURLWithPath: "/usr/bin/git")
```
BrandonWeng: I dont think we can assume everyone has this actually. Non technical users may not have git installed? Unless all Macs come with this pre bundled

Sources/FluidAudio/DownloadUtils.swift
```diff
@@ -1,289 +1,175 @@
...
+            let errorMessage = String(data: errorData, encoding: .utf8) ?? "Unknown error"
+            logger.error("❌ Git clone failed: \(errorMessage)")
+            throw URLError(.cannotConnectToHost, userInfo: [
+                NSLocalizedDescriptionKey: "Failed to download repository: \(errorMessage)"
+            ])
         }
 
-        try data.write(to: destinationPath)
-    }
+        // Check if Git LFS files need to be pulled
```
BrandonWeng: https://github.com/huggingface/swift-transformers/blob/main/Sources/Hub/Downloader.swift

Might be able to refer to this @Alex-Wengg

Sources/FluidAudio/DownloadUtils.swift
```diff
@@ -1,289 +1,202 @@
...
+        
+        // Check if git is available
+        let gitPath = "/usr/bin/git"
+        guard FileManager.default.fileExists(atPath: gitPath) else {
+            logger.error("❌ Git not found at \(gitPath)")
+            throw URLError(.cannotFindHost, userInfo: [
+                NSLocalizedDescriptionKey: "Git is not installed. Please install Xcode Command Line Tools by running 'xcode-select --install' in Terminal."
+            ])
         }
+
```
BrandonWeng: Wait no I don think this will work. Majority of users will not have this installed

PR #43, opened 2025-07-29
Add Voice dataset (subset) benchmark for vad

BrandonWeng:
MUSAN is mostly used for training, it would be better to benchmark against something like VOiCES dataset. Added a command to test and run it on a small subset (50). 

Updated the Github workflow for VAD as well.


Alex-Wengg:
the VAD Benchmark Results having 100%  Accuracy rate is a big sus. And shouldn't we include a RTF or RTFx row too . what about comparing against other datasets

PR #44, opened 2025-07-29
initial setup

Alex-Wengg:


Sources/FluidAudioCLI/Commands/OpenBenchBenchmark.swift
```diff
@@ -0,0 +1,457 @@
...
+    static func run(arguments: [String]) async {
+        let benchmarkStartTime = Date()
+        
+        // Configuration
+        var datasets: [String] = []
+        var maxFiles: Int? = nil
+        var outputFile = "openbench_results.json"
+        var threshold: Float = 0.7
+        var debugMode = false
+        var openBenchPath = "/Users/kikow/brandon/OpenBench"
```
BrandonWeng, line 16: ?

BrandonWeng, line 16: ? 

Would this be better included in your openbench fork instead of FluidAudio?

Issue #53, opened 2025-08-01
Load function ignores user specified computeUnits


Steven-Weng:
**Bug**:
I noticed that the `AsrModels.load(from:configuration:)` ignores the caller-supplied `configuration.computeUnits` and always uses `optimizedConfig.computeUnits`.
Current implementation:
```
// AsrModels.swift
public static func load(
        from directory: URL,
        configuration: MLModelConfiguration? = nil     
    ) async throws -> AsrModels {

        let config = configuration ?? defaultConfiguration()  
        …
        for (modelName, modelType) in modelConfigs {
            let optimizedConfig = optimizedConfiguration(for: modelType)

            _ = try await DownloadUtils.loadModels(
                .parakeet,
                modelNames: [modelName],
                directory: directory.deletingLastPathComponent(),
                computeUnits: optimizedConfig.computeUnits        // <-- optimizedConfig is always chosen
            )
        }
}
```

BrandonWeng:
Hmm good catch, but is there a particular reason why you need to customize the configs?

Steven-Weng:
I need to customize the configs to use CPU and Neural Engine so my iOS app can run background transcription. Apple doesn't let you use the GPU if your app is backgrounded:
```
IOGPUMetalError: Insufficient Permission (to submit GPU work from background) (00000006:kIOGPUCommandBufferCallbackErrorBackgroundExecutionNotPermitted)
Execution of the command buffer was aborted due to an error during execution. Insufficient Permission (to submit GPU work from background) (00000006:kIOGPUCommandBufferCallbackErrorBackgroundExecutionNotPermitted)
IOGPUMetalError: Insufficient Permission (to submit GPU work from background) (00000006:kIOGPUCommandBufferCallbackErrorBackgroundExecutionNotPermitted)
[Espresso::handle_ex_plan] exception=Espresso exception: “Generic error”: Insufficient Permission (to submit GPU work from background) (00000006:kIOGPUCommandBufferCallbackErrorBackgroundExecutionNotPermitted); code=7 status=-1
Unable to compute the prediction using a neural network model. It can be an invalid input data or broken/unsupported model (error code: -1).
```

BrandonWeng:
Should be fixed by the PR!

PR #54, opened 2025-08-01
Fix ComputeUnits override for ASR

BrandonWeng:
https://github.com/FluidInference/FluidAudio/issues/53

Ideally users should be able to configure each model but honestly from our benchmark tests, anything other than ANE was tested and the performance is not ideal

### DIFF
--- a/Sources/FluidAudio/ASR/ANEOptimizer.swift
+++ b/Sources/FluidAudio/ASR/ANEOptimizer.swift
@@ -94,25 +94,8 @@ public enum ANEOptimizer {
 
     /// Configure optimal compute units for each model type
     public static func optimalComputeUnits(for modelType: ModelType) -> MLComputeUnits {
-        switch modelType {
-        case .melSpectrogram:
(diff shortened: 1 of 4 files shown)


Sources/FluidAudio/ASR/ANEOptimizer.swift
```diff
@@ -94,25 +94,8 @@ public enum ANEOptimizer {
...
-            // Small dense layers perfect for Neural Engine
-            if #available(macOS 14.0, iOS 17.0, *) {
-                // Use Neural Engine exclusively for small models
-                return .all  // Will prefer NE for small ops
-            } else {
-                return .cpuAndNeuralEngine
-            }
-        }
+        // Testing shows CPU+ANE is fastest for all models, including melspectrogram
+        return .cpuAndNeuralEngine
```
BrandonWeng, line 98: GPU for MelSpec turned out to not be faster, up to 111x RTFx when I run locally on this change

BrandonWeng:
@claude review

PR #58, opened 2025-08-03
fix: vocabulary loading ignores custom model directory

Beingpax:
AsrManager.loadVocabulary() used hardcoded paths, ignoring custom directories specified in AsrModels.downloadAndLoad(to: customDirectory). This would cause the transcription to fail when using custom directory configurations. 
CleanShot 2025-08-03 at 09 22 46@2x

**Fix:** Move vocabulary loading into AsrModels.load() to respect the same directory path used for ML models.

### DIFF
--- a/Sources/FluidAudio/ASR/AsrManager.swift
+++ b/Sources/FluidAudio/ASR/AsrManager.swift
@@ -269,7 +269,6 @@ public final class AsrManager {
         decoderState = freshState
     }
 
-
     private func loadModel(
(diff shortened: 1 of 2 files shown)


BrandonWeng (commented):
thanks for the change! 

Could you run formato nthe code please

```bash
swift format --in-place --recursive --configuration .swift-format Sources/ 
```

BrandonWeng (approved):

Issue #74, opened 2025-08-20
Compile Errors on iOS.


Steven-Weng:
It seems that some recent changes introduced code that isn’t compatible with iOS. After updating the package, FluidAudio no longer compiles, and the following errors appear in `DownloadUtils`:
```
kCFNetworkProxiesHTTPSEnable' is unavailable in iOS
kCFNetworkProxiesHTTPSProxy' is unavailable in iOS
kCFNetworkProxiesHTTPSPort' is unavailable in iOS
```

BrandonWeng:
Thanks! Seems related to this: https://github.com/FluidInference/FluidAudio/pull/72

BrandonWeng:
https://github.com/FluidInference/FluidAudio/pull/75

BrandonWeng:
Should be fixed on main now!

PR #75, opened 2025-08-20
Only allow proxy for macOS and add test for iOS test with xCode build

BrandonWeng:
[breaks builds for iOS](https://github.com/FluidInference/FluidAudio/issues/74)

### What does this change do?
Only build the proxy components for macOS, not iOS

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,24 @@
+# FluidAudio - Agent Development Guide
+
+## Build & Test Commands
+```bash
+swift build                                    # Build project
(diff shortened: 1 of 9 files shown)


AGENTS.md
```diff
@@ -0,0 +1,24 @@
+# FluidAudio - Agent Development Guide
```
BrandonWeng, line 1: Supports other agnetic coding tools like opencode and cursor for other contributors

BrandonWeng:
Just gonna merge this. Very simple change

PR #86, opened 2025-08-29
Async post processing for ASR benchmarks

BrandonWeng:
Speed up the benchmarks a little bit at the end by making it async, probably not a ton but for something we run so often every little bit helps


BrandonWeng:
```
2620 files per dataset • Test runtime: 3m 41s • 08/28/2025, 11:08 PM EDT
--- Benchmark Results ---
   Dataset: librispeech test-clean
   Files processed: 2620
   Average WER: 3.2%
   Median WER: 0.0%
   Average CER: 1.5%
   Median RTFx: 99.7x
   Overall RTFx: 110.3x (19452.5s / 176.4s)
```

BrandonWeng:
Nevermind, looked at the duration diff and there's no difference lol

PR #96, opened 2025-09-09
Fix: adjust confidence array during streaming token deduplication

amby:
Fix array length mismatch bug in transcribeStreamingChunk where the confidence array was not properly adjusted when duplicate tokens were removed, causing processTranscriptionResult to receive inconsistent array lengths.

Problem:
- When duplicate tokens were removed from streaming chunks, only timestamps were adjusted by dropping the first removedCount elements
- The confidence array remained unchanged, creating a length mismatch
- This caused createTokenTimings to fail silently and return empty arrays
- Array lengths: confidences.count > tokenIds.count > timestamps.count

Solution:
- Add adjustedConfidences calculation that mirrors adjustedTimestamps logic
- Drop first removedCount elements from confidence array when duplicates removed
- Ensure all arrays (tokens, timestamps, confidences) maintain consistent lengths

Impact:
- Fixes token timing creation in streaming ASR scenarios
- Prevents silent failures in processTranscriptionResult
- Aligns streaming deduplication with ChunkProcessor implementation
- Maintains backward compatibility

Tests: All existing tests pass (AsrManagerExtensionTests, AsrTranscriptionTests, StreamingAsrManagerTests)

### DIFF
--- a/Sources/FluidAudio/ASR/AsrManager.swift
+++ b/Sources/FluidAudio/ASR/AsrManager.swift
@@ -10,7 +10,7 @@ public enum AudioSource: Sendable {
 @available(macOS 13.0, *)
 public final class AsrManager {
 
-    internal let logger = Logger(subsystem: "com.fluidinfluence.asr", category: "ASR")
+    internal let logger = AppLogger(category: "ASR")
     internal let config: ASRConfig
 
     internal var melspectrogramModel: MLModel?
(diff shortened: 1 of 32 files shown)


BrandonWeng:
Makes sense - could you run `swift format --in-place --recursive --configuration .swift-format ./` 

We're reworking streaming here as well so no need for additonal tests

BrandonWeng (approved):

amby:
Sure thing. Ran `swift format --in-place --recursive --configuration .swift-format ./` and committed the change.

PR #97, opened 2025-09-09
Unified logger for CLI commands too

BrandonWeng:
Avoid the annoying "print" when developing and testing the CLI. Also making sure we don't miss things in the logger.error calls when running in the CLI 

`swift build` + integration tests should pass

### DIFF
--- a/Sources/FluidAudio/Shared/AppLogger.swift
+++ b/Sources/FluidAudio/Shared/AppLogger.swift
@@ -70,9 +70,10 @@ public struct AppLogger {
     }
 
     // MARK: - Console Mirroring
-
     private func logToConsole(_ level: Level, _ message: String) {
-        Task { await LogConsole.shared.write(level: level, category: category, message: message) }
+        Task.detached(priority: .utility) {
+            await LogConsole.shared.write(level: level, category: category, message: message)
+        }
     }
 }
 


Sources/FluidAudio/ASR/AsrModels.swift
```diff
@@ -15,7 +15,7 @@ public struct AsrModels: Sendable {
     public let configuration: MLModelConfiguration
     public let vocabulary: [Int: String]
 
-    private static let logger = Logger(subsystem: "com.fluidinfluence.asr", category: "AsrModels")
```
BrandonWeng, line 18: lol typo

BrandonWeng:
> ### ⚠️ Issues to Address
> #### 1. **Performance Concern (Blocker)**
> **File**: `Sources/FluidAudio/Shared/AppLogger.swift:66`
> 
> The `logToConsole` method creates a new `Task` for every log message, which could lead to significant overhead in high-frequency logging:
> 
> ```swift
> // Current - creates task explosion risk
> private func logToConsole(_ level: Level, _ message: String) {
>     Task { await LogConsole.shared.write(level: level, category: category, message: message) }
> }
> 
> // Suggested fix
> private func logToConsole(_ level: Level, _ message: String) {
>     Task.detached(priority: .utility) {
>         await LogConsole.shared.write(level: level, category: category, message: message)
>     }
> }
> ```

True

Issue #100, opened 2025-09-11
Need cocoapods support


rpatnayakuni22:
Our project still only supports cocoapods and I tried using SPM but seeing errors on CI. Is it possible to add cocoapods support? Or if anyone has done any work to consume this via cocoapods please share. Thanks in advance!

BrandonWeng:
You might be able to use 

https://github.com/trinhngocthuyen/cocoapods-spm

@claude can you confirm and provide an example of the config

rpatnayakuni22:
Thank you so much for the quick response @BrandonWeng 
Can I do spm_pkg in .podspcec? As I'm first integrating this into a library/framework and then adding to the App. 
spm_pkg 'FluidAudio', 
          url: 'https://github.com/FluidInference/FluidAudio.git', 
          requirement: .branch('main')

BrandonWeng:
> Thank you so much for the quick response [@BrandonWeng](https://github.com/BrandonWeng) Can I do spm_pkg in .podspcec? As I'm first integrating this into a library/framework and then adding to the App. spm_pkg 'FluidAudio', url: 'https://github.com/FluidInference/FluidAudio.git', requirement: .branch('main')

Based on the examples, I think you need to add it in the Podfile, with cocoapods-spm, I don't think we need a .podspec file  

https://github.com/trinhngocthuyen/cocoapods-spm/blob/main/examples/Podfile

Btw - just curious, are you adding it to an existing project or a new one? If its a new project, why cocoapods over SPM?

rpatnayakuni22:
It's an old existing project. I have to add it to a framework first (that's where I have to add to podspec) and then the app will use podspec and app's podfile to successfully build everything!

BrandonWeng:
Here, give this a try :). We don't have any cocoa projects so please do let us know if you run into issues! 

```
 🎉  Congrats

 🚀  FluidAudio (0.4.1) successfully published
 📅  September 11th, 16:11
 🌎  https://cocoapods.org/pods/FluidAudio
 👍  Tell your friends!
```

https://github.com/FluidInference/FluidAudio/pull/102

rpatnayakuni22:
@BrandonWeng Thank you so much for the quick PR!! will let you know once I try!!

BrandonWeng:
Feel free to re-open or open another issue, going to close this issue! 

If you want to debug together, welcome to ping me on Discord.

PR #102, opened 2025-09-11
.podspec for cocoapod support

BrandonWeng:

We don't have any external deps so its not hard to support podspec, see the issue for more details. 

https://github.com/FluidInference/FluidAudio/issues/100

cocoapod-spm works for apps with Podfiles but some frameworks won't have that so they need podspec

### DIFF
--- a/FluidAudio.podspec
+++ b/FluidAudio.podspec
@@ -19,6 +19,6 @@ Pod::Spec.new do |spec|
   spec.source       = { :git => "https://github.com/FluidInference/FluidAudio.git", :tag => "v#{spec.version}" }
   spec.source_files = "Sources/FluidAudio/**/*.swift"
   
-  spec.swift_versions = ["5.10"]
+  spec.swift_versions = ["5.10", "6.0"]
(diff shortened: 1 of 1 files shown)


FluidAudio.podspec
```diff
@@ -0,0 +1,24 @@
+Pod::Spec.new do |spec|
+  spec.name         = "FluidAudio"
+  spec.version      = "0.4.1"
+  spec.summary      = "Speaker diarization, voice-activity-detection and transcription with CoreML"
```
Alex-Wengg, line 4: what is this . never heard of cocoapod before, is this some docker alt https://github.com/FluidInference/FluidAudio/issues/100#issue-3407109043

BrandonWeng: no, its an older way of publishing swift/objective-c packages for iOS app. Older apps still use it, but most have moved to Swift Package Manager (spm)

Alex-Wengg: what the hack

PR #132, opened 2025-10-02
Add token-level timestamps to StreamingTranscriptionUpdate

### LINKED ISSUE #120: StreamingASRManager Token level timestamps for diarization
Currently, the StreamingASRManager only returns segments of text in StreamingTranscriptionUpdate but not the individual tokens and their associated timestamps.

If we want to be able to diarize with a token level precision, it sounds like we would need that information. Do you think we could add token level timestamps to the StreamingTranscriptionUpdate object? What do you think?

BrandonWeng:
Closes #120

Adds token-level timestamps to StreamingTranscriptionUpdate for token-level precision in diarization.

## Changes
- Add tokenTimings property to StreamingTranscriptionUpdate struct
- Update StreamingAsrManager to pass tokenTimings from ASRResult
- Change finish() method to return ASRResult instead of String
- Update CLI commands to handle new finish() return type
- Add test for StreamingTranscriptionUpdate with tokenTimings

This enables token-level precision for diarization by exposing individual token timestamps from the ASR model.

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
+++ b/Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
@@ -267,7 +267,8 @@ public actor StreamingAsrManager {
         let minimumProcessingTime: TimeInterval = 1e-6
 
         let maxTimestampFrame = accumulatedTokenTimestamps.max() ?? 0
-        let derivedSampleCount = maxTimestampFrame > 0
+        let derivedSampleCount =
+            maxTimestampFrame > 0
             ? (maxTimestampFrame + 1) * ASRConstants.samplesPerEncoderFrame : 0
         let sampleCount = max(totalSamplesProcessed, derivedSampleCount)
         let processingTime = sampleCount > 0 ? max(elapsedTime, minimumProcessingTime) : elapsedTime


BrandonWeng:
@codex take another pass

Sources/FluidAudio/ASR/AsrTranscription.swift
```diff
@@ -189,11 +189,13 @@ extension AsrManager {
...
         processingTime: TimeInterval,
-        tokenTimings: [TokenTiming] = []
+        tokenTimings: [TokenTiming] = [],
+        totalSampleCount: Int? = nil
     ) -> ASRResult {
 
         let (text, finalTimings) = convertTokensWithExistingTimings(tokenIds, timings: tokenTimings)
-        let duration = TimeInterval(audioSamples.count) / TimeInterval(config.sampleRate)
+        let effectiveSampleCount = totalSampleCount ?? audioSamples.count
+        let duration = TimeInterval(effectiveSampleCount) / TimeInterval(config.sampleRate)
```
Alex-Wengg, line 198: would we ever need a cover for divided by 0

BrandonWeng: samplke rate has to be > 0, it really iss always 16000

Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
```diff
@@ -314,9 +407,12 @@ public actor StreamingAsrManager {
                 previousTokens: accumulatedTokens
             )
 
+            let frameOffset = calculateGlobalFrameOffset(for: windowStartSample)
+            let adjustedTimestamps = timestamps.map { $0 + frameOffset }
```
Alex-Wengg, line 456: what would be the point of frameoffsets

Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
```diff
@@ -219,12 +266,46 @@ public actor StreamingAsrManager {
...
+    }
+
+    func accumulateTokenMetadata(tokens: [Int], timestamps: [Int], confidences: [Float]) {
+        guard !tokens.isEmpty else { return }
+
+        accumulatedTokens.append(contentsOf: tokens)
+
+        if timestamps.count == tokens.count {
+            accumulatedTokenTimestamps.append(contentsOf: timestamps)
+        } else if !timestamps.isEmpty {
```
Alex-Wengg: what about an else statement case

Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
```diff
@@ -458,6 +555,18 @@ public actor StreamingAsrManager {
     }
 }
 
+#if DEBUG
```
Alex-Wengg, line 603: is this needed

Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
```diff
@@ -161,25 +165,67 @@ public actor StreamingAsrManager {
...
+                totalSampleCount: finalSampleCount
             )
-            finalText = finalResult.text
         } else {
             // Fallback to text concatenation if no tokens available
-            finalText = confirmedTranscript + volatileTranscript
+            let fallbackText = confirmedTranscript + volatileTranscript
+            finalResult = ASRResult(
+                text: fallbackText,
+                confidence: 1.0,
```
Alex-Wengg: why would the confidence fallback be 1.0 ? here

Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
```diff
@@ -299,7 +388,11 @@ public actor StreamingAsrManager {
     }
 
     /// Process a single assembled window: [left, chunk, right]
-    private func processWindow(_ windowSamples: [Float], actualLeftSeconds: Double) async {
+    private func processWindow(
+        _ windowSamples: [Float],
+        actualLeftSeconds _: Double,
```
Alex-Wengg: is this being used

Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
BrandonWeng: honestly true, we shouldn't even need this else

Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
BrandonWeng: for tests

Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
BrandonWeng: for the global timestamp, it needs to be adjusted

Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
BrandonWeng: not showing in the diff but this was just a format change. 

https://github.com/FluidInference/FluidAudio/blob/e144f8247a296de5813d51f810eca5d5993e5e7f/Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift#L373

BrandonWeng: oh true, poroabbly not used in the actual function

BrandonWeng: @claude  remove the unused variable

Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
```diff
@@ -219,12 +213,149 @@ public actor StreamingAsrManager {
...
+
+        return (sampleCount, processingTime)
+    }
+
+    private func buildFinalResult(processingTime: TimeInterval, sampleCount: Int) -> ASRResult {
+        guard let asrManager = asrManager, !accumulatedTokens.isEmpty else {
+            let duration = TimeInterval(sampleCount) / TimeInterval(config.asrConfig.sampleRate)
+            return ASRResult(
+                text: finalTranscriptText(),
+                confidence: 1.0,
```
Alex-Wengg, line 284: why is this hardcoded to 1.0

Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
```diff
@@ -219,12 +213,149 @@ public actor StreamingAsrManager {
...
+    private func resolveFinalMetrics(at now: Date) -> (sampleCount: Int, processingTime: TimeInterval) {
+        let elapsedTime = startTime.map { now.timeIntervalSince($0) } ?? 0
+        let minimumProcessingTime: TimeInterval = 1e-6
+
+        let maxTimestampFrame = accumulatedTokenTimestamps.max() ?? 0
+        let derivedSampleCount =
+            maxTimestampFrame > 0
+            ? (maxTimestampFrame + 1) * ASRConstants.samplesPerEncoderFrame : 0
+        let sampleCount = max(totalSamplesProcessed, derivedSampleCount)
+        let processingTime = sampleCount > 0 ? max(elapsedTime, minimumProcessingTime) : elapsedTime
```
Alex-Wengg, line 274: whats the difference between these two variables exactly

Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
```diff
@@ -219,12 +213,149 @@ public actor StreamingAsrManager {
...
+        label: String,
+        omissionDetail: String
+    ) -> [Value] {
+        guard !values.isEmpty else { return [] }
+
+        guard values.count == expectedCount else {
+            logger.warning(
+                "Final token \(label) count (\(values.count)) does not match token count (\(expectedCount)); \(omissionDetail)"
+            )
+            return []
```
Alex-Wengg, line 334: why are we returning [] for a mismatch count ?

Sources/FluidAudio/ASR/Streaming/StreamingAsrManager.swift
```diff
@@ -458,6 +600,18 @@ public actor StreamingAsrManager {
     }
 }
 
+#if DEBUG
+extension StreamingAsrManager {
+    internal func setAsrManagerForTesting(_ manager: AsrManager?) {
```
Alex-Wengg, line 605: should this be async ?

Alex-Wengg:
comments done

BrandonWeng:
Not happy with the PR, going to rewrite

PR #151, opened 2025-10-20
Fix Kokoro again.. File phonic not found issue with xFramework

BrandonWeng:

https://github.com/user-attachments/assets/0d62b276-d0c6-4972-a243-f9a40fe768fa

Builds and runs on iOS properly now, using croqeteer's test app ^ with kokoro + vad + mlx


Alex-Wengg (approved):
Current -> A , bruh what is this shit

PR #150, opened 2025-10-20
pyannote community-1 model for offline speaker diarization pipeline

BrandonWeng:
### Why is this change needed?

Keeping the streaming one around as the VBx and AHC clustering gets pretty expensive after 30mins of audio and running it constantly gets expensive. Its still possible to support clustering between files but will save that for another PR. 

Pyannote's Bench mark is around 11%  - i increased steps to 0.2s instead of 0.1 to double the speed but also selective fp16 results in more operations to run on ANE but also means that we lose some precision. 

```
Average DER: 14.95% | Median DER: 10.89% | Average JER: 39.27% | Median JER: 40.74% (collar=0.25s, ignoreOverlap=True)
Average RTFx: 139.63 (from 232 clips)
Metrics summary saved to: /Users/brandonweng/FluidAudioDatasets/voxconverse/metrics/test_metrics_release.json
Completed. New results: 232, Skipped existing: 0, Total attempted: 232
```

See benchmark.md for more info but compared to Pytorch model, we are 100x faster than the CPU version and ~6x faster compared to the mps backend on mb pro 4

PR #152, opened 2025-10-21
Clean up unneeded version checks

BrandonWeng:

Pulling some of the changes from this PR here to break it down into smaller PRs. 

https://github.com/FluidInference/FluidAudio/pull/150

### DIFF
--- a/Sources/FluidAudio/ASR/TDT/BlasIndex.swift
+++ b/Sources/FluidAudio/ASR/TDT/BlasIndex.swift
@@ -1,0 +1,15 @@
+import Accelerate
+
+#if ACCELERATE_NEW_LAPACK
+typealias BlasIndex = __LAPACK_int
+#else
+typealias BlasIndex = Int32
+#endif
+
+@inline(__always)
+func makeBlasIndex(_ value: Int, label: String) throws -> BlasIndex {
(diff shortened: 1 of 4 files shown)


Sources/FluidAudio/ASR/TDT/EncoderFrameView.swift
```diff
@@ -76,29 +76,13 @@ struct EncoderFrameView {
         guard hiddenStride != 0 else {
             throw ASRError.processingFailed("Invalid hidden stride: 0")
         }
-        guard let elementCount = Int32(exactly: hiddenSize) else {
```
BrandonWeng, line 79: from strwaming PR

Sources/FluidAudio/ASR/TDT/BlasIndex.swift
```diff
@@ -0,0 +1,15 @@
+import Accelerate
+
+#if ACCELERATE_NEW_LAPACK
+typealias BlasIndex = __LAPACK_int
+#else
+typealias BlasIndex = Int32
+#endif
+
+@inline(__always)
```
BrandonWeng, line 9: the accelerate flag only works for us to stop seeing the warning when building, still a problem when people don't set the flags in xcode

PR #153, opened 2025-10-21
consistent use of passed-in config

starcrest:
Otherwise the start condition is evaluated with one config and the end/hysteresis with another.

### DIFF
--- a/Documentation/API.md
+++ b/Documentation/API.md
@@ -52,11 +52,11 @@ Voice activity detection using the Silero VAD Core ML model with 256 ms unified 
 - `VadManager.sampleRate = 16000`
 
 **Configuration (`VadConfig`):**
-- `threshold: Float` — Decision threshold (0.0–1.0). Default: `0.85`.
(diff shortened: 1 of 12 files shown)


BrandonWeng:
@claude can you look at the build error 

```
ork/FluidAudio/FluidAudio/Sources/FluidAudio/VAD/VadManager+Streaming.swift:44:32: error: value of type 'VadSegmentationConfig' has no member 'threshold'
 42 |         nextState.processedSamples += chunkSampleCount
 43 | 
 44 |         let threshold = config.threshold
    |                                `- error: value of type 'VadSegmentationConfig' has no member 'threshold'
 45 |         let negativeThreshold = config.effectiveNegativeThreshold(baseThreshold: threshold)
 46 |         let speechPadSamples = Int(config.speechPadding * Double(Self.sampleRate))
[9/70] Compiling FluidAudio LexiconAssetManager.swift
/Users/runner/work/FluidAudio/FluidAudio/Sources/FluidAudio/VAD/VadManager+Streaming.swift:44:32: error: value of type 'VadSegmentationConfig' has no member 'threshold'
 42 |         nextState.processedSamples += chunkSampleCount
 43 | 

```

PR #172, opened 2025-10-31
Make ANE Utils concurrency safe

BrandonWeng:

Unit test to reproduce what the user was seeing, but essentially has to do with shared ML arrays in the ANE optimizer when we have multiple models running at once, or when there's too many instances running (meeting note taker for instance) 

Previously we had applied fixes on the manager level but the culpruit is the underlying ane optimizer so it manifested in different forms

### DIFF
--- a/Sources/FluidAudio/Shared/ANEMemoryOptimizer.swift
+++ b/Sources/FluidAudio/Shared/ANEMemoryOptimizer.swift
@@ -1,81 +1,7 @@
 import Accelerate
 import CoreML
-import Dispatch
 import Foundation
 import Metal
-
-private final class ANEBufferPool {
-
-    private struct Entry {
-        var array: MLMultiArray
-        var leased: Bool
-    }
-
-    private var storage: [String: [Entry]] = [:]
-    private let queue = DispatchQueue(label: "com.fluidaudio.ane-buffer-pool")
-
-    func sharedBuffer(
-        key: String,
-        shape: [NSNumber],
-        dataType: MLMultiArrayDataType,
-        allocator: () throws -> MLMultiArray
-    ) throws -> MLMultiArray {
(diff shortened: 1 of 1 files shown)


Sources/FluidAudio/Diarizer/Core/DiarizerManager.swift
```diff
@@ -342,6 +346,47 @@ public final class DiarizerManager {
         return (segments, timings)
     }
 
+    private func extractEmbeddingsSynchronously<C>(
```
BrandonWeng: a wrapper that is needed because we don't want to surface teh async/await to the end user.

Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
```diff
@@ -3,38 +3,14 @@ import CoreML
...
 
 /// Embedding extractor with ANE-aligned memory and zero-copy operations
-public class EmbeddingExtractor {
+public final class EmbeddingExtractor {
     private let wespeakerModel: MLModel
     private let logger = AppLogger(category: "EmbeddingExtractor")
     private let memoryOptimizer = ANEMemoryOptimizer()
 
-    // Pre-allocated ANE-aligned buffers
-    private var waveformBuffer: MLMultiArray?
```
BrandonWeng, line 12: no impact and a lot of concurrency issues arise from this


main RTFx `305.2` this branch RTFx `299.5`

BrandonWeng: The shared pool is what i suspect to be causing the conurrency issues wand thread crashing

Alex-Wengg (approved):

PR #177, opened 2025-11-03
Switch ASR to stateless for batching 

BrandonWeng:

Stateless would make streaming easier and it seems to improve the WER for v2 and v3, which is really surprising. 

Before:
```
Metrics
 WER               9.05%
 CER               7.94%
 Reference Words   12766
 Hypothesis Words  12234
```

After:
```text
Metrics
 WER               4.01%
 CER               3.01%
 Reference Words   12766
 Hypothesis Words  12666
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ Make a PR if you want to add your app, please keep it in chronological order.
 | **[mac-whisper-speedtest](https://github.com/anvanvan/mac-whisper-speedtest)** | Comparison of different local ASR, including one of the first verions of FluidAudio's ASR models |
 | **[Starling](https://github.com/Ryandonofrio3/Starling)** | Open Source, fully local voice-to-text transcription with auto-paste at your cursor. |
 | **[BoltAI](https://boltai.com/)** | Write content 10x faster using parakeet models |
+| **[Voxeoflow](https://www.voxeoflow.app)** | Mac dictation app with real-time translation. Lightning-fast transcription in over 100 languages, instantly translated to your target language. |
 
 ## Installation
 


Alex-Wengg:
> Stateless would make streaming easier and it seems to improve the WER for v2 and v3, which is really surprising.


isn't this pr for chunnking not streaming ?

Sources/FluidAudio/ASR/ChunkProcessor.swift
```diff
@@ -6,85 +6,127 @@ struct ChunkProcessor {
...
-            }
-
             // Combine tokens, timestamps, and confidences into aligned tuples
             guard windowTokens.count == windowTimestamps.count && windowTokens.count == windowConfidences.count else {
                 throw ASRError.processingFailed("Token, timestamp, and confidence arrays are misaligned")
             }
 
-            let windowData = zip(zip(windowTokens, windowTimestamps), windowConfidences).map {
+            let windowData: [TokenWindow] = zip(zip(windowTokens, windowTimestamps), windowConfidences).map {
                 (token: $0.0.0, timestamp: $0.0.1, confidence: $0.1)
```
Alex-Wengg, line 73: do we ever use confidence in our chunker , i know in the streaming part it is used

BrandonWeng:
> > Stateless would make streaming easier and it seems to improve the WER for v2 and v3, which is really surprising.
> 
> isn't this pr for chunnking not streaming ?

I meant later when we do update teh stremaing imeplemntnation

Sources/FluidAudio/ASR/ChunkProcessor.swift
BrandonWeng: may not be but some people like to have it in the result. Im testing confidence based merging but so far it seems to make things worse

Sources/FluidAudio/ASR/ChunkProcessor.swift
```diff
@@ -6,85 +6,127 @@ struct ChunkProcessor {
...
+            )
+        }
 
-            centerStart += centerSamples
+        if chunkOutputs.count > 1 {
+            let decode: (TokenWindow) -> String = { window in
+                manager.vocabulary[window.token] ?? "<\(window.token)>"
+            }
+            for chunk in chunkOutputs.dropFirst() {
+                mergedTokens = mergeChunks(mergedTokens, chunk, decode: decode)
```
Alex-Wengg: should we ignore Punctuation tokens for this function ?

Sources/FluidAudio/ASR/ChunkProcessor.swift
```diff
@@ -96,108 +138,284 @@ struct ChunkProcessor {
...
+                matches: contiguousPairs,
+                overlapLeft: overlapLeft,
+                overlapRight: overlapRight,
+                left: left,
+                right: right,
+                decode: decode
+            )
+        }
+
+        let lcsPairs = findLongestCommonSubsequencePairs(
```
Alex-Wengg, line 214: wouldn't mergeUsingMatches handle like 90% of cases . i recall the ASR python uses LCS here

Sources/FluidAudio/ASR/ChunkProcessor.swift
BrandonWeng: No, could be merging near puncuations too

Alex-Wengg:
> > > Stateless would make streaming easier and it seems to improve the WER for v2 and v3, which is really surprising.
> > 
> > 
> > isn't this pr for chunnking not streaming ?
> 
> I meant later when we do update teh stremaing imeplemntnation

i guess bc we are using overlapping chunks here, the decoder model & tdt algo were not trained on for overlapping so the overlapping tokens must have greatly confused the decoder model

Sources/FluidAudio/ASR/ChunkProcessor.swift
BrandonWeng: honestly collecting different fallbacks from different implementations to try and reduce WER overall and this combo seems to result in the best results

BrandonWeng:
> > > > Stateless would make streaming easier and it seems to improve the WER for v2 and v3, which is really surprising.
> > > 
> > > 
> > > isn't this pr for chunnking not streaming ?
> > 
> > 
> > I meant later when we do update teh stremaing imeplemntnation
> 
> i guess bc we are using overlapping chunks here, the decoder model & tdt algo were not trained on for overlapping so the overlapping tokens must have greatly confused the decoder model

not entirely - the nemo implementation actually uses a sliding window. It might just be how the decoder algorithm is designed, it speeds things up quite a bit but also results in potentially inocrrectly skipping text too

Alex-Wengg (approved):
i guess decoders were not trained for overlapping perhaps

PR #182, opened 2025-11-06
add @available(iOS 17.0, *) for VBxClustering and OfflineEmbeddingExt…

Josscii:
Although right now it dropped iOS 16 support, with this change, users can still change the package.swift to work on iOS 16.

### DIFF
--- a/Sources/FluidAudio/Diarizer/Offline/Clustering/VBxClustering.swift
+++ b/Sources/FluidAudio/Diarizer/Offline/Clustering/VBxClustering.swift
@@ -22,7 +22,7 @@ import os.signpost
 ///
 /// The implementation includes warm-start initialization from initial hard cluster
 /// assignments and supports PLDA whitening transformation of input features.
-@available(iOS 17.0, *)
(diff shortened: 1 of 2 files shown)


BrandonWeng:
Thanks for the PR. Why do you only need to annotate these two classes? Could you leave a comment

Josscii:
I would like to change the Package.swift to support iOS 16:
```
    platforms: [
        .macOS(.v13),
        .iOS(.v16),
    ],
```
and found that, the only errors I encounter is these two missing annotatations, cause it use `PLDATransform` which also annotated, so  if I add these, it complied correctly.

BrandonWeng:
> I would like to change the Package.swift to support iOS 16:
> 
> ```
>     platforms: [
>         .macOS(.v13),
>         .iOS(.v16),
>     ],
> ```
> 
> and found that, the only errors I encounter is these two missing annotatations, cause it use `PLDATransform` which also annotated, so if I add these, it complied correctly.

Alright, sounds good to me. Thanks for the PR!

PR #186, opened 2025-11-15
optionalize TTS via FluidAudioTTS target

Alex-Wengg:
- Make TTS optional to avoid GPL by default; enable with `FLUIDAUDIO_ENABLE_TTS=1.`
- Split TTS into FluidAudioTTS target; CLI imports it only when enabled.
- Expose MLModel.compatPrediction as public for cross‑module use.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -13,8 +13,8 @@ let package = Package(
             targets: ["FluidAudio"]
         ),
         .library(
-            name: "FluidAudioWithTTS",
-            targets: ["FluidAudio", "FluidAudioTTS"]
+            name: "FluidAudioTTS",
+            targets: ["FluidAudioTTS"]
         ),
         .executable(
             name: "fluidaudio",
--- a/README.md
+++ b/README.md
@@ -65,11 +65,33 @@ dependencies: [
 ],
 ```
 
+### Choosing a Product
+
+FluidAudio provides two library products:
+
+- **`FluidAudio`** (default) - Core functionality: ASR, diarization, VAD
+  - Lightweight, no GPL dependencies
+  - Recommended for most apps
+
+- **`FluidAudioTTS`** - Text-to-Speech (Kokoro)
+  - Separate optional product
+  - Includes ESpeakNG framework (GPL-3.0)
+  - Only bundled if you explicitly add it as a dependency
+
+**In Xcode:**
+1. Add the FluidAudio package to your project
+2. In the "Add Package" dialog, select your desired product(s):
+   - `FluidAudio` for core features (ASR, diarization, VAD)
+   - `FluidAudioTTS` if you need text-to-speech
+3. Add the selected product(s) to your app target
+
 **In Package.swift:**
 ```swift
+// Core features only (no GPL dependencies):
 .product(name: "FluidAudio", package: "FluidAudio")
-// OR for TTS support:
-.product(name: "FluidAudioWithTTS", package: "FluidAudio")
+
+// Add TTS support (includes GPL ESpeakNG):
+.product(name: "FluidAudioTTS", package: "FluidAudio")
 ```
 
 **In Xcode:**


.github/workflows/gemini-pr-review.yml
```diff
@@ -13,13 +13,19 @@ on:
...
-      (github.event_name == 'issue_comment' && 
-       github.event.issue.pull_request && 
-       contains(github.event.comment.body, '/fluid review') &&
-       (github.event.comment.author_association == 'OWNER' ||
-        github.event.comment.author_association == 'MEMBER' ||
-        github.event.comment.author_association == 'COLLABORATOR'))
+      (
+        github.event_name == 'workflow_dispatch' ||
+        (github.event_name == 'issue_comment' && 
+         github.event.issue.pull_request &&
```
BrandonWeng: lets jsut remove gemini pr review

Package.swift
```diff
@@ -1,5 +1,8 @@
 // swift-tools-version: 5.10
 import PackageDescription
+import Foundation
+
+let enableTTS = ProcessInfo.processInfo.environment["FLUIDAUDIO_ENABLE_TTS"] == "1"
```
BrandonWeng: hmm, do we not need to update instructions in readme and stuff?

FluidAudio.podspec
```diff
@@ -17,7 +17,9 @@ Pod::Spec.new do |spec|
   spec.osx.deployment_target = "14.0"
 
   spec.source       = { :git => "https://github.com/FluidInference/FluidAudio.git", :tag => "v#{spec.version}" }
-  spec.swift_versions = ["5.10"]
+  # CocoaPods sets SWIFT_VERSION based on this list; use values Xcode recognizes.
+  # Our code remains compatible with Swift 5 toolchains (uses conditional compilation for 6.x APIs).
+  spec.swift_versions = ["5.9"]
```
BrandonWeng: ? why does this change

README.md
```diff
@@ -434,6 +434,18 @@ Requirements (macOS)
 Ensure eSpeak NG headers/libs are available via pkg-config (`espeak-ng`).
 <https://github.com/espeak-ng/espeak-ng/tree/master>
 
+### Enable TTS (build flag)
+
+TTS is an optional component. Enable it at build time with the `FLUIDAUDIO_ENABLE_TTS=1` environment variable:
```
BrandonWeng: add reason here  why..

README.md
```diff
@@ -435,6 +435,24 @@ Requirements (macOS)
...
+TTS is disabled by default to keep the core library lightweight and license‑clean for most apps. Enabling it pulls in the ESpeakNG framework (used for G2P/OOV handling) and the TTS targets, which:
+
+- Avoids shipping GPL‑compatible bits unless you explicitly opt in
+- Keeps default builds smaller and CI faster when TTS isn’t needed
+- Limits platform scope (ESpeakNG is supported and linked on macOS in this repo)
+
+Enable TTS at build time with the `FLUIDAUDIO_ENABLE_TTS=1` environment variable:
+
+- Terminal (SwiftPM):
+  - `FLUIDAUDIO_ENABLE_TTS=1 swift run fluidaudio tts "Hello" --output out.wav`
```
BrandonWeng: how would this work in xcode btw, most of the people using TTS are via XCode, Also what does MLX audio for kokoro do here?

Alex-Wengg: i separated into FluidAudioWithTTS and FluidAudio products

Package.swift
```diff
@@ -12,44 +15,81 @@ let package = Package(
             name: "FluidAudio",
             targets: ["FluidAudio"]
         ),
+        .library(
+            name: "FluidAudioWithTTS",
+            targets: ["FluidAudio", "FluidAudioTTS"]
+        ),
```
BrandonWeng, line 18: why do we need the env if there's a sepearte library with the TTS target?

Alex-Wengg: good point removed

Alex-Wengg:
> made edits to the readme myself.. can you look at the changes i made.
> 
> "product" doesn't really make sense here... as a user its really confusing because there's "library" in the code.
> 
> btw, i still see that you have binary taget in the spec, how do we know if it actually gets bundled or not?

good catch, addressed

PR #191, opened 2025-11-19
Streaming Diarization Improvements

SGD2718:
- Fixed `SegmentProcessor.getSegments` internal access level
- Added audio conversion method for `CMSampleBuffer`
- Added a `startTime` argument to `DiarizerManager.performCompleteDiarization` to offset segment timestamps.
- Added `AudioStream` struct to simplify timestamp calculations for a streamed/online audio input source with a lot of overlap

### DIFF
--- a/Sources/FluidAudio/Shared/AudioConverter.swift
+++ b/Sources/FluidAudio/Shared/AudioConverter.swift
@@ -76,7 +76,7 @@ final public class AudioConverter {
     /// - Returns: Float array at 16kHz mono
     /// - Throws: `AudioConverterError.sampleBufferFormatMissing` (most likely caused by the sample buffer belonging
     ///  to a video frame)
-    public func convertSampleBuffer(_ sampleBuffer: CMSampleBuffer) throws -> [Float] {
+    public func resampleSampleBuffer(_ sampleBuffer: CMSampleBuffer) throws -> [Float] {
         let buffer = try extractAVAudioPCMBuffer(from: sampleBuffer)
         return try convertBuffer(buffer, to: targetFormat)
     }


Alex-Wengg:
@SGD2718 might be useful to add a doc for how to do streaming mode in diarizer. also if you could find any benchmarks for it too .

SGD2718:
It appears that my change to the CLI thing broke it

edit: nvm i messed up a merge

SGD2718:
Summary of changes:
  - Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
      - Adds embeddingWindowSampleCount, uses pooled ANE-aligned buffers (getPooledBuffer) instead of per-call allocation.
      - Copies audio once, prefetches waveform once, reuses a single MLPredictionOptions, and reorders masking to avoid repeated waveform copies.
      - Adds repeatVoiceSegmentIfNeeded to tile shorter audio within the fixed embedding window.
  - Sources/FluidAudio/Diarizer/Segmentation/SegmentationProcessor.swift
      - Adds repeatAudioChunk(_:chunkSize:buffer:) to zero-pad and repeat audio slices to the chunk size while preserving time base.
      - Uses vectorized copies/clears and repeats via vDSP.
  - Sources/FluidAudio/Diarizer/Core/DiarizerManager.swift
      - Calls repeatAudioChunk on the padded chunk before embedding extraction; passes the repeated chunk to the extractor while masks still come from the original
        padded chunk.
  - Sources/FluidAudioCLI/Commands/ProcessCommand.swift
      - Adds CLI flags --chunk-duration and --chunk-overlap, and passes them into DiarizerConfig for streaming mode.

SGD2718:
Update: there are some issues with the diariation fix.

Alex-Wengg:
Will take look tmw @SGD2718 can you add details to the PR body,  something like this https://github.com/FluidInference/FluidAudio/pull/191#issuecomment-3554496096

Documentation/SpeakerDiarization.md
```diff
@@ -1,3 +1,4 @@
+
 # Speaker Diarization
 
 Real-time speaker diarization for iOS and macOS, answering "who spoke when" in audio streams.
```
Alex-Wengg: how real time we talking, can i do a 1 second chunking or what ? . a general subjective guidline. generally 0-5 seconds is considered "real-time" , ~ 10 seconds chunk is considered "near real time"

Documentation/SpeakerDiarization.md
```diff
@@ -255,35 +256,24 @@ Process audio in chunks for real-time applications:
...
-            let adjustedSegment = TimedSpeakerSegment(
-                speakerId: segment.speakerId,
-                startTimeSeconds: streamPosition + segment.startTimeSeconds,
-                endTimeSeconds: streamPosition + segment.endTimeSeconds
-            )
-            handleSpeakerSegment(adjustedSegment)
-        }
+var stream = AudioStream(
+    chunkDuration: 5.0, // 10.0 for best accuracy. 3.0 for lowest
+    strideDuration: 2.0, // Duration between successive chunk starts
```
Alex-Wengg: this line isn't very clear , can we try another variable instead of  strideDuration

Documentation/SpeakerDiarization.md
```diff
@@ -255,35 +256,24 @@ Process audio in chunks for real-time applications:
...
+    atTime: 0.0, // Audio stream start time
```
Alex-Wengg: why ?

Documentation/SpeakerDiarization.md
```diff
@@ -322,18 +312,30 @@ import AVFoundation
...
     init() async throws {
         let models = try await DiarizerModels.downloadIfNeeded()
         diarizer = DiarizerManager()  // Default config
         diarizer.initialize(models: models)
+        audioStream = AudioStream(
+            chunkDuration: 5.0,
+            strideDuration: 3.0,
+            atTime: 0.0,
+            alignment: .frontAligned,
+            processGaps: Bool = false
```
Alex-Wengg: why do we need this boolean at all

Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
```diff
@@ -51,6 +52,29 @@ public final class EmbeddingExtractor {
         // to maintain compatibility with the rest of the pipeline
         var embeddings: [[Float]] = []
 
+        let audioSampleCount = audio.distance(from: audio.startIndex, to: audio.endIndex)
+        let effectiveSampleCount = min(audioSampleCount, waveformSampleCount)
```
Alex-Wengg: effectiveSampleCount ? why not sampleCount. whats the difference between audioSampleCount and effectiveSampleCount .

Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
```diff
@@ -62,18 +86,12 @@ public final class EmbeddingExtractor {
                 continue
             }
```
Alex-Wengg: why is this removed

Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
```diff
@@ -106,28 +124,60 @@ public final class EmbeddingExtractor {
         return embeddings
     }
 
+    private func fillWaveformBuffer<C>(
```
Alex-Wengg, line 114: not very intuitive , maybe paddingAudio.

Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
```diff
@@ -106,28 +124,60 @@ public final class EmbeddingExtractor {
...
+        var filled = copyCount
+        while filled < targetSampleCount {
+            let repeatCount = min(filled, targetSampleCount - filled)
+            let destination = ptr.advanced(by: filled)
+            memmove(destination, ptr, repeatCount * MemoryLayout<Float>.size)
+            filled += repeatCount
+        }
+    }
+
     private func fillMaskBufferOptimized(
```
Alex-Wengg: fillWaveformBuffer vs fillMaskBufferOptimized ?

Sources/FluidAudio/Diarizer/Segmentation/SegmentationProcessor.swift
```diff
@@ -11,6 +11,81 @@ public struct SegmentationProcessor {
 
     public init() {}
 
+    /// Repeat audio to fill the expected chunk length for downstream models.
+    ///
+    /// This keeps timestamps stable because only the audio is repeated; segment timing
```
Alex-Wengg: these comments can be worded better

Sources/FluidAudio/Shared/AudioConverter.swift
```diff
@@ -57,6 +58,46 @@ final public class AudioConverter {
         return try resampleBuffer(buffer)
     }
 
+    /// Convert a CMSampleBuffer to 16kHz mono Float32 samples
```
Alex-Wengg: i feel like this is already done before , deja vu . are you able to verify

Sources/FluidAudio/Shared/AudioConverter.swift
```diff
@@ -292,6 +336,12 @@ public enum AudioConverterError: LocalizedError {
             return "Failed to create conversion buffer"
         case .conversionFailed(let error):
             return "Audio conversion failed: \(error?.localizedDescription ?? "Unknown error")"
+        case .sampleBufferFormatMissing:
+            return "Sample buffer is missing a valid audio format description"
```
Alex-Wengg: why would we need this ??

Sources/FluidAudio/Shared/AudioConverter.swift
```diff
@@ -292,6 +336,12 @@ public enum AudioConverterError: LocalizedError {
...
+        case .frameCountOverflow:
+            return "Sample buffer contains more frames than supported"
+        case .sampleBufferCopyFailed(let status):
```
Alex-Wengg: what could cause this error though

Alex-Wengg (commented):
first impression this looks pretty decent, i will review more in detail after 


are you able to add a full diarizer benchmark for near real time to benchmark.md ?

Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
SGD2718: The changes in the embeddingExtractor were made with Codex. I figured that they worked since the benchmark didn't indicate a noticeable slowdown, while the accuracy did improve.

Documentation/SpeakerDiarization.md
SGD2718: The idea is that if the user pauses the app and the clock continues moving forward, then it will simply skip the long gap of silence when it resumes, instead of processing the silence. Although, in hindsight, I guess that could also be accomplished by leaving the time argument for the ScoreStream.write function blank.

Documentation/SpeakerDiarization.md
SGD2718: The atTime argument is there so the stream may be aligned relative to a global clock instead of the start of the audio.

Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
Alex-Wengg: spend some time thinking about variable names that are more intuitive .

Documentation/SpeakerDiarization.md
Alex-Wengg: think about which stra is better then

SGD2718:
what happened to the ASR and VAD benchmarks i never even touched those

SGD2718:
Summary of changes:
- Added `AudioStream` struct to facilitate timestamping for real-time audio streams with overlapping chunks
- DiarizerManager.performCompleteDiarization() now has a startTime parameter to remove the need for manual timestamp adjustments while streaming
- Made embedding extractor repeat-pad the audio and segment masks for chunks shorter than 10s
- Changed SegmentationProcessor.getSegments() from an internal to a public method
- Updated DiarizationManager.md to show a real-time diarization implementation using the AudioStream class instead of manual timestamping.
- Added missing command for real-time diarization benchmarks in benchmark.md

Alex-Wengg:
https://github.com/actions/runner-images/issues/13341 we are currently exping a weird bug with github i have attempted means to get around the issue but so far no success

SGD2718:
Summary of changes:
- Added `AudioStream` struct to facilitate timestamping for real-time audio streams with overlapping chunks
- `DiarizerManager.performCompleteDiarization()` now has a `startTime` parameter to remove the need for manual timestamp adjustments while streaming
- `EmbeddingExtractor.getEmbeddings` now loops the audio signal for chunks shorter than 10s
- Changed `SegmentationProcessor.getSegments()` access level from `internal` to a `public`
- Updated `DiarizationManager.md` to show a real-time diarization implementation using the AudioStream class instead of manual timestamping.
- Added missing command for real-time diarization benchmarks in `Benchmark.md`
- `ANEMemoryOptimizer.optimizedCopy` now performs a move for sources of type `ContiguousArray<Float>`
- The `process` CLI command now has arguments for `chunk-seconds` and `overlap-seconds`.
- The `diarization-benchmark` command now uses an `autoreleasepool` to avoid crashing when there are a lot of chunks

Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
```diff
@@ -35,6 +35,7 @@ public final class EmbeddingExtractor {
         }
 
         let waveformShape = [3, 160_000] as [NSNumber]
+        let waveformSampleCount = waveformShape.last?.intValue ?? 160_000
```
Alex-Wengg: why this tertiary specifically

Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
```diff
@@ -51,6 +52,17 @@ public final class EmbeddingExtractor {
         // to maintain compatibility with the rest of the pipeline
         var embeddings: [[Float]] = []
 
+        // Fill shared waveform buffer once; reused across speakers
+        fillWaveformBufferOptimized(
```
Alex-Wengg: fillWaveformBufferOptimized -> fillWaveformBuffer

Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
```diff
@@ -106,8 +112,54 @@ public final class EmbeddingExtractor {
         return embeddings
     }
 
+    /// Fill the waveform buffer with loop (repeat) padding
+    private func fillWaveformBufferOptimized<C>(
```
Alex-Wengg: this feels convoluted, can you see it can be simplified or compacted

Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
```diff
@@ -119,15 +171,35 @@ public final class EmbeddingExtractor {
...
                 ptr,
-                vDSP_Length(maskCount),
+                vDSP_Length(numMasksUsed),
+                vDSP_Length(1),
+                vDSP_Length(1),
+                vDSP_Length(numMasksUsed)
+            )
+        }
+
+        guard numMasksUsed < maskCount && numMasksUsed > 0 else {
```
Alex-Wengg: is this necessary

Sources/FluidAudio/Shared/AudioConverter.swift
Alex-Wengg: can you verify this is even needed ? i recall we have something like this before

Sources/FluidAudio/Shared/AudioConverter.swift
```diff
@@ -292,6 +336,12 @@ public enum AudioConverterError: LocalizedError {
...
+            return "Failed to copy samples from CMSampleBuffer (status: \(status))"
```
Alex-Wengg, line 365: what edge case would cause this error to pop up

Sources/FluidAudio/Shared/AudioConverter.swift
```diff
@@ -292,6 +336,12 @@ public enum AudioConverterError: LocalizedError {
             return "Failed to create conversion buffer"
         case .conversionFailed(let error):
             return "Audio conversion failed: \(error?.localizedDescription ?? "Unknown error")"
+        case .sampleBufferFormatMissing:
+            return "Sample buffer is missing a valid audio format description"
+        case .frameCountOverflow:
```
Alex-Wengg: i am. just confused by how sampleBufferFormatMissing, frameCountOverflow and sampleBufferCopyFailed error might happen , i recall we have rail guards against these edge cases

Sources/FluidAudio/Shared/AudioStream.swift
```diff
@@ -0,0 +1,387 @@
...
+    ///   - time: Timestamp for resynchronization (optional)
+    private mutating func writeContiguousSamples<C>(from source: C, atTime time: TimeInterval? = nil) throws
+    where C: Collection & Sequence, C.Element == Float {
+        guard source.count > 0 else {
+            return
+        }
+
+        let writeIndexReset = (chunkingStrategy == .useMostRecent ? temporaryChunkSize : buffer.count)
+
+        if let time {
```
Alex-Wengg: this logic is hard to follow & quite complex . consider if can be simplified down

Sources/FluidAudio/Shared/AudioStream.swift
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    /// Add new audio data to the buffer
+    /// - Parameters:
+    ///   - source: Audio samples to write
+    ///   - time: Timestamp for resynchronization (optional)
+    /// - Warning: Samples may be skipped if the time jumps forward significantly.
+    public mutating func write<C>(from source: C, atTime time: TimeInterval? = nil) throws
+    where C: Collection & Sequence, C.Element == Float {
+        if let array = source as? [Float] {
```
Alex-Wengg, line 208: do these type checkings really help ? its an internal library so its very unlikely for the data types to be altered

Sources/FluidAudio/Shared/AudioStream.swift
```diff
@@ -0,0 +1,387 @@
...
+
+    /// Prepare the buffer to append from a buffer to avoid overflowing
+    /// - Parameters:
+    ///   - src: Pointer to the start of the source buffer
+    ///   - count: Number of samples about to be appended
+    ///   - maxWriteIndex: Maximum value of `writeIndex` to avoid shifting back the buffer
+    ///   - shiftedWriteIndex: Desired value of `writeIndex` after shifting back the buffer
+    /// - Returns: Pointer to the first element of the source buffer that will actually be appended
+    @discardableResult
+    private mutating func prepareToAppend(
```
Alex-Wengg: - Mutates count parameter (inout) - unusual and confusing
- Complex control flow with multiple exit points
- Hard to reason about correctness

Sources/FluidAudio/Shared/AudioStream.swift
```diff
@@ -0,0 +1,387 @@
...
+    public init(
+        chunkDuration: TimeInterval = 10.0,
+        chunkSkip: TimeInterval? = nil,
+        streamStartTime time: TimeInterval = 0.0,
+        chunkingStrategy: AudioStreamChunkingStrategy = .useMostRecent,
+        startupStrategy: AudioStreamStartupStrategy = .startSilent,
+        sampleRate: Double = 16_000,
+        bufferCapacitySeconds: TimeInterval? = nil
+    ) throws {
+        guard chunkDuration >= 1 / sampleRate else {
```
Alex-Wengg: why 1/sampleRate specifically

Sources/FluidAudio/Shared/AudioStream.swift
```diff
@@ -0,0 +1,387 @@
...
+
+            if missingSampleCount > 0 {
+                prepareToAppend(
+                    from: nil,
+                    count: &missingSampleCount,
+                    maxWriteIndex: buffer.count - source.count,
+                    shiftedWriteIndex: writeIndexReset - source.count)
+                if missingSampleCount > 0 {
+                    let stride = MemoryLayout<Float>.stride
+                    memset(&buffer[writeIndex], 0, missingSampleCount * stride)
```
Alex-Wengg: this might be dangerous if we are gonna use this consider possible guard rails

Sources/FluidAudio/Shared/AudioStream.swift
```diff
@@ -0,0 +1,387 @@
...
+                    writeIndex += missingSampleCount
+                }
+            } else if missingSampleCount < 0 {
+                rollbackNewest(-missingSampleCount)
+            }
+        }
+
+        let success: Void? = source.withContiguousStorageIfAvailable { ptr in
```
Alex-Wengg: doesn't the code already handle this , also this naming might be misleading it could be assumed to be related to audio buffer

Tests/FluidAudioTests/AudioStreamTests.swift
Alex-Wengg, line 1: i believe we have a ASR STT streaming algo unit tests you can take inspirations from , i think the biggest issue we faced was related to timestamp syncing

Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
SGD2718: Residue from Codex that i missed

Sources/FluidAudio/Shared/AudioStream.swift
SGD2718: so that there will be at least one sample in there

Sources/FluidAudio/Shared/AudioStream.swift
SGD2718: yeah I wasn't a big fan of the way i did this but I was trying to remove redundancies.

The idea is that the algorithm is something like this:

If the buffer is about to overflow from adding the new samples, then shift everything back to make room.
If the source exceeds the size of the buffer, then cut it short to fit it in a chunk (for useMostRecent chunking) or to fit in the buffer (for useFixedStep).

Sources/FluidAudio/Shared/AudioStream.swift
SGD2718: prepareToAppend will cut missingSampleCount short if it's about to overflow.

Sources/FluidAudio/Shared/AudioConverter.swift
SGD2718: I think you're getting it mixed up with the one that converts the AVAudioPCMBuffer. Converting from a CMSampleBuffer is a bit different and is used when the audio stream comes from a live video feed, which has both microphone and video data.

Sources/FluidAudio/Shared/AudioConverter.swift
SGD2718: They sampleBufferFormatMissing will occur if someone sends in a video sampleBuffer by mistake. As for the other one, I will remove that.

Sources/FluidAudio/Shared/AudioStream.swift
SGD2718: Yes. It's a public method in a public class.

Sources/FluidAudio/Shared/AudioStream.swift
SGD2718: The logic used to set writeIndexReset was moved to append() and appendZeros().

Sources/FluidAudio/Shared/AudioConverter.swift
SGD2718: For sampleBufferCopyFailed:
```swift
// AAC format description
var asbd = AudioStreamBasicDescription(
    mSampleRate: 44100,
    mFormatID: kAudioFormatLinearPCM,
    mFormatFlags: kAudioFormatFlagIsSignedInteger | kAudioFormatFlagIsPacked,
    mBytesPerPacket: 2,
    mFramesPerPacket: 1,
    mBytesPerFrame: 2,
    mChannelsPerFrame: 1,
    mBitsPerChannel: 16,
    mReserved: 0
)

let asbdPtr = withUnsafePointer(to: &asbd) { $0 }

var desc: CMAudioFormatDescription?
CMAudioFormatDescriptionCreate(
    allocator: nil,
    asbd: asbdPtr,
    layoutSize: 0,
    layout: nil,
    magicCookieSize: 0,
    magicCookie: nil,
    extensions: nil,
    formatDescriptionOut: &desc
)

// Allocate a block buffer *smaller* than what PCM needs
var blockBuffer: CMBlockBuffer?
CMBlockBufferCreateWithMemoryBlock(
    allocator: nil,
    memoryBlock: nil,
    blockLength: 1,   // ❌ needs at least 2 bytes for 1 sample
    blockAllocator: nil,
    customBlockSource: nil,
    offsetToData: 0,
    dataLength: 1,
    flags: 0,
    blockBufferOut: &blockBuffer
)

var sampleBuffer: CMSampleBuffer?
CMSampleBufferCreate(
    allocator: nil,
    dataBuffer: blockBuffer,
    dataReady: true,
    makeDataReadyCallback: nil,
    refcon: nil,
    formatDescription: desc!,
    sampleCount: 1,
    sampleTimingEntryCount: 0,
    sampleTimingArray: nil,
    sampleSizeEntryCount: 0,
    sampleSizeArray: nil,
    sampleBufferOut: &sampleBuffer
)

guard let sampleBuffer else {
    return
}
let converter = AudioConverter()

do {
    _ = try converter.resampleSampleBuffer(sampleBuffer)
} catch {
    print(error)
}
```

This will print 

```
sampleBufferCopyFailed(-12704)
```

SGD2718: As for why anyone in their right mind would do this, I don't know.

Sources/FluidAudio/Shared/AudioStream.swift
SGD2718: I have changed this to be more clear by instead making the guard statement check if `chunkSize > 0`

Tests/FluidAudioTests/AudioStreamTests.swift
SGD2718: Added some test cases to ensure timestamps work

Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
```diff
@@ -118,7 +160,9 @@ public final class EmbeddingExtractor {
         vDSP_vfill(&zero, ptr, 1, vDSP_Length(totalElements))
 
         // Copy speaker mask to first slot using optimized memory copy
-        let maskCount = masks[speakerIndex].count
+        let requiredMasks = masks[speakerIndex].count
```
Alex-Wengg: shouldn't this be the other way , 

his is backwards. You're setting maskCount (current size) to numMasksUsed (which is already the calculated target size), then trying to pad up to masks[speakerIndex].count. The logic should be:
```
let requiredMasks = numMasksUsed  // Target size to pad to
var maskCount = min(masks[speakerIndex].count, numMasksUsed)  // Current size
```

Sources/FluidAudio/Shared/AudioConverter.swift
```diff
@@ -292,6 +336,12 @@ public enum AudioConverterError: LocalizedError {
             return "Failed to create conversion buffer"
         case .conversionFailed(let error):
             return "Audio conversion failed: \(error?.localizedDescription ?? "Unknown error")"
+        case .sampleBufferFormatMissing:
+            return "Sample buffer is missing a valid audio format description. Make sure it's not a video sample."
```
Alex-Wengg: `            return "Sample buffer is missing a valid audio format description. Make sure it's not a video sample." 
            `
can we make this more formal , we can add a code comment above this line ' something edge case for video audio sample may have different audio formats'

Sources/FluidAudio/Shared/AudioConverter.swift
Alex-Wengg: did. you encounter this issue when you did streaming mode

Sources/FluidAudio/Shared/AudioStream.swift
```diff
@@ -0,0 +1,432 @@
+import AVFoundation
+import Foundation
+
+/// A sliding window buffer for a real-time audio stream.
+public class AudioStream {
```
Alex-Wengg: thoughts on making this an actor class to make it thread safe ?

Sources/FluidAudio/Shared/AudioStream.swift
```diff
@@ -0,0 +1,432 @@
...
+    /// Bind a callback to the chunk updates
+    /// - Parameter callback: The callback to bind
+    public func bind(_ callback: @escaping (ArraySlice<Float>, TimeInterval) throws -> Void) {
+        self.callback = callback
+    }
+
+    /// Bind a callback to the chunk updates
+    /// - Parameter callback: The callback to bind
+    public func bind(_ callback: @escaping (ArraySlice<Float>, TimeInterval) async throws -> Void) {
+        self.callback = { (chunk: ArraySlice<Float>, timestamp: TimeInterval) -> Void in
```
Alex-Wengg: do you think its worth adding a error callback ?

Sources/FluidAudio/Shared/AudioStream.swift
SGD2718: If an audio stream is being passed around across different tasks instead of having it's chunks being copied and passed around, then that probably means it was designed poorly. It would make more sense to have a central external manager that grabs the chunks and deploys them across concurrent branches

Sources/FluidAudio/Diarizer/Extraction/EmbeddingExtractor.swift
SGD2718: No. `numMasksUsed` is the number of masks used in the chunk. All elements of `masks` always have the same length (specifically, 589 masks) regardless of the length of the chunk. I will rename `numMasksUsed` to `numMasksInChunk` for more clarity.

Sources/FluidAudio/Shared/AudioStream.swift
SGD2718: Regarding the `sampleBufferCopyFailed` error, it's still possible to trigger it if someone decides to make their own `CMSampleBuffer` and does it wrong. I provided a code example earlier that would achieve it. While it's basically impossible for most of the errors to occur under normal usage, I still think that it's better to throw instead of silently letting it pass in the event that someone does do something funny

Sources/FluidAudio/Shared/AudioStream.swift
SGD2718: I think it makes more sense to just remove the `throws`

Alex-Wengg: how come though

Sources/FluidAudio/Shared/AudioStream.swift
Alex-Wengg: is this class design for 
- Single consumer
- Multiple read-only consumers
- Multiple read-write consumers

Sources/FluidAudio/Shared/AudioStream.swift
SGD2718: because if i add an error callback then the method will take two arguments and then users can't do this anymore:
```swift
audioStream.bind { chunk, time in
    await foo(chunk, time)
}
```
The error callback would likely come after the callback argument.

Sources/FluidAudio/Shared/AudioStream.swift
SGD2718: It's designed for one reader and one writer because the chunks are popped as soon as they're read. However, I guess that it would make sense to have the reader and writer be asynchronous. I will add a session queue so that it also works in synchronous contexts easily.

Alex-Wengg: alright lmk when you made the changes. i will approve right after

SGD2718:
- AudioStream is now Sendable.
- AudioStream chunks are now `[Float]`s because the callback was moved to execute after the read to avoid blocking the session queue for extended periods of time
- AudioStream bindings now accept generic collections for the chunks
- I also made it possible to silence the AudioConverter's debug logs because they were getting extremely annoying when I was trying to do real-time diarization tests

Alex-Wengg (approved):

PR #199, opened 2025-11-23
Make DiarizerManager Sendable with internal locking

saamerm:
Problem:
Swift 6 strict concurrency checking prevents DiarizerManager from being passed to background tasks (e.g., Task.detached) because it is a class that does not conform to Sendable. This makes it difficult to run diarization inference off the Main Actor without warnings or errors.

Solution:

- Marked DiarizerManager as @unchecked Sendable.
- Introduced OSAllocatedUnfairLock to protect mutable state (models and embeddingExtractor). 
- Added backing storage (_models, _embeddingExtractor) to support the locking mechanism. 
- Updated cleanup() to properly nullify the embedding extractor as well. 

Safety Justification:
Thread safety is achieved because the public API captures a local reference to the models and embeddingExtractor inside the lock before beginning long-running inference tasks. If cleanup() is called concurrently, the inference method holds onto the valid reference until it completes, preventing race conditions on the pointer.

### DIFF
--- a/Sources/FluidAudio/ASR/AsrTranscription.swift
+++ b/Sources/FluidAudio/ASR/AsrTranscription.swift
@@ -140,7 +140,7 @@ extension AsrManager {
 
     /// Streaming-friendly chunk transcription that preserves decoder state and supports start-frame offset.
     /// This is used by both sliding window chunking and streaming paths to unify behavior.
-    internal func transcribeStreamingChunk(
+    public func transcribeStreamingChunk(
         _ chunkSamples: [Float],
         source: AudioSource,
         previousTokens: [Int] = []
--- a/Sources/FluidAudio/Frameworks/ESpeakNG.xcframework/Info.plist
+++ b/Sources/FluidAudio/Frameworks/ESpeakNG.xcframework/Info.plist
@@ -6,23 +6,11 @@
 	<array>
 		<dict>
 			<key>BinaryPath</key>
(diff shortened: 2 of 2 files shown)


Alex-Wengg:
https://github.com/actions/runner-images/issues/13341 we are currently exping a weird bug with github job runner i have attempted means to get around the issue but so far no success

saamerm:
@Alex-Wengg no worries, does that mean no merges till that is resolved?

Alex-Wengg:
@saamerm they reesolved it , i will take a look

Alex-Wengg:
just a headups we aren't on swift 6 as of this moment but we do plan to migrate soon

saamerm:
@Alex-Wengg ah I see. I was trying to use it in the sample repo yall have “swift-scribe” which runs swift 6, so I had manually pulled the files into that project. This PR helps allow better usage away from the UI thread without workarounds

Alex-Wengg:
we might plan to migrate to swift 6 soon

xinnjie:
Better to use Mutex in the new `Synchronization` library rather than the old lock.  Note that `Synchronization` requires IOS 18.

Or use make DiarizerManager an actor. But API will change.

Alex-Wengg:
hi @saamerm we have just migrated over to swift 6 so this change may not be needed anymore.

saamerm:
@Alex-Wengg sounds good, thank you

PR #210, opened 2025-12-08
Add Intel Mac detection and model validation utilities

Alex-Wengg:
## Summary
- `SystemInfo.isAppleSilicon` and `SystemInfo.isIntelMac` to detect platform
-  `AsrModels.isModelValid()` validates all 4 Parakeet components (Preprocessor, Encoder, Decoder, Joint) can load without corruption
- Reuse decoder state arrays to prevent memory accumulation during streaming
- Handle non-contiguous strides in copyData

### What VoiceInk should do (app side)

| Issue | VoiceInk Fix |
|-------|-------------|
| Intel Mac users selecting Parakeet | Use `SystemInfo.isIntelMac` to hide/disable Parakeet models in UI |
| Infinite "Transcribing" hang | Add timeout to transcription calls with user-facing error |
| 20-30s delay after sleep | Show "Loading model..." UI during model load (ANE recompilation is Apple's `anecompilerservice`, cannot be sped up) |
| Model corruption | Use `AsrModels.isModelValid()` before transcription, prompt re-download if invalid |


Sources/FluidAudio/ASR/AsrModelWarmup.swift
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    /// Check current warmup status
+    ///
+    /// Returns a status object with information about model readiness
+    public func checkStatus(version: AsrModelVersion = .v3) -> WarmupStatus {
+        let cacheValid = AsrModels.isModelCacheValid(version: version)
+        let aneCompiling = AsrModels.isANECompilerRunning()
+        let isIntel = AsrModels.isIntelMac
```
BrandonWeng: we dont support intel though. im pretty sure the package explicitly doens't allow intels

Sources/FluidAudio/ASR/AsrModelWarmup.swift
```diff
@@ -0,0 +1,163 @@
...
+/// Usage:
+/// ```swift
+/// // Start listening for wake events (call once at app startup)
+/// AsrModelWarmup.shared.startListening()
+///
+/// // Or manually trigger a warmup
+/// await AsrModelWarmup.shared.warmupModels()
+/// ```
+@available(macOS 10.15, iOS 13.0, *)
+public final class AsrModelWarmup: @unchecked Sendable {
```
BrandonWeng: i thought we already have warm ups already wwhen intiailizing the asr manager?

Sources/FluidAudio/DownloadUtils.swift
```diff
@@ -186,20 +186,58 @@ public class DownloadUtils {
         public static let `default` = DownloadConfig()
     }
 
+    /// Model loading configuration
+    public struct LoadConfig: Sendable {
```
BrandonWeng: you wanna just switch to the swift hugging face officilial client? it hsould handle al ot of the errors

Alex-Wengg: might be better in another PR

BrandonWeng: yeah but that or could go in first

BrandonWeng: then you don’t need these changes

Sources/FluidAudio/ASR/AsrManager.swift
```diff
@@ -78,6 +78,11 @@ public final class AsrManager {
     /// Initialize ASR Manager with pre-loaded models
     /// - Parameter models: Pre-loaded ASR models
     public func initialize(models: AsrModels) async throws {
+        #if arch(x86_64)
+        logger.warning(
+            "Intel Mac detected - FluidAudio requires Apple Silicon. Transcription will likely fail.")
+        #endif
```
BrandonWeng: this doesn't really do anything.. also where is the complaint about intel coming from? IRRC we have docs saying that intel devices are not supported. 

this isn't fixing anything....

Sources/FluidAudio/Shared/SystemInfo.swift
```diff
@@ -7,6 +7,27 @@ import Darwin
 
 /// System information utilities and one-time environment reporting.
 public enum SystemInfo {
+    // MARK: - Architecture detection
+
+    /// Returns `true` if running on Apple Silicon (ARM64).
+    /// Use this to guard features that require Neural Engine (ANE) support.
```
BrandonWeng: a lot of these comments are pretty redundant

Sources/FluidAudio/ASR/AsrModels.swift
```diff
@@ -370,6 +370,58 @@ extension AsrModels {
...
+    ///   - version: The model version to validate (default: .v3)
+    /// - Returns: `true` if models exist and all components load successfully
+    ///
+    /// - Note: This does NOT check ANE cache state. First load after system sleep
+    ///   may still take 20+ seconds as `anecompilerservice` recompiles for Neural Engine.
+    public static func isModelValid(version: AsrModelVersion = .v3) async -> Bool {
+        // Intel Macs cannot run Parakeet models
+        guard SystemInfo.isAppleSilicon else {
+            logger.warning("Model validation failed: Intel Mac detected, Parakeet requires Apple Silicon")
+            return false
```
BrandonWeng: this should just throw hard or fail?

Sources/FluidAudio/ASR/TDT/TdtDecoderState.swift
```diff
@@ -51,8 +51,14 @@ struct TdtDecoderState {
     }
 
     mutating func update(from decoderOutput: MLFeatureProvider) {
-        hiddenState = decoderOutput.featureValue(for: "h_out")?.multiArrayValue ?? hiddenState
-        cellState = decoderOutput.featureValue(for: "c_out")?.multiArrayValue ?? cellState
+        // Copy data into existing arrays instead of replacing them to avoid memory leaks.
```
BrandonWeng: should we jus t remove the memory optimziation and see how much worse the perf is? irrc its not a huge performance

Sources/FluidAudio/ASR/AsrModels.swift
Alex-Wengg: i think it should be something like the dev themselves should decide. returning false might be easier and they can add a throw fail if they want

BrandonWeng: but in this case its just failing silently? i dont htink the coreml models would even compiled acutally

Alex-Wengg: now throws error instead of returning false then

Sources/FluidAudio/ASR/TDT/TdtDecoderState.swift
BrandonWeng: what about RTFx? latency

BrandonWeng: this should not affect WER

Alex-Wengg: RTFx is about the same after running 1000 files on the main vs this branch

BrandonWeng (approved):
lgtm but remove json files pls

PR #213, opened 2025-12-12
add support for custom vocabulary

smdesai:
Custom lexicons exist to let users override pronunciation for domain-specific terminology that general-purpose G2P and built-in dictionaries either do not support of mishandle —this is especially common in finance (tickers, company names, jargon like “EBITDA”), healthcare (drug names, procedures, acronyms), legal (Latinisms, case citations), and tech (product names, abbreviations). The goal is for these overrides to be dependable and to take effect exactly when the user expects resulting in fluid audio generation.

### DIFF
--- a/Sources/FluidAudio/ASR/AsrModels.swift
+++ b/Sources/FluidAudio/ASR/AsrModels.swift
@@ -370,6 +370,43 @@ extension AsrModels {
         return modelsPresent && vocabPresent
     }
 
+    /// Validates that Parakeet models exist and can be loaded.
+    /// Throws on Intel Mac since CoreML models require Apple Silicon.
+    public static func isModelValid(version: AsrModelVersion = .v3) async throws -> Bool {
+        guard SystemInfo.isAppleSilicon else {
+            throw ASRError.unsupportedPlatform("Parakeet models require Apple Silicon")
+        }
+
+        let cacheDir = defaultCacheDirectory(for: version)
+        guard modelsExist(at: cacheDir, version: version) else {
+            logger.info("Model validation failed: model files not found")
+            return false
+        }
+
+        let repoPath = repoPath(from: cacheDir, version: version)
+        let config = MLModelConfiguration()
+        config.computeUnits = .cpuOnly
+
+        let modelsToValidate = [
+            ("Preprocessor", ModelNames.ASR.preprocessorFile),
+            ("Encoder", ModelNames.ASR.encoderFile),
+            ("Decoder", ModelNames.ASR.decoderFile),
+            ("Joint", ModelNames.ASR.jointFile),
+        ]
+
+        for (name, fileName) in modelsToValidate {
+            let modelPath = repoPath.appendingPathComponent(fileName)
+            do {
+                _ = try MLModel(contentsOf: modelPath, configuration: config)
+            } catch {
+                logger.error("Model validation failed: \(name) - \(error.localizedDescription)")
+                return false
+            }
+        }
+
+        return true
+    }
+
     public static func defaultCacheDirectory(for version: AsrModelVersion = .v3) -> URL {
         let appSupport = FileManager.default.urls(
             for: .applicationSupportDirectory, in: .userDomainMask
--- a/Sources/FluidAudio/ASR/AsrTypes.swift
+++ b/Sources/FluidAudio/ASR/AsrTypes.swift
@@ -73,6 +73,7 @@ public enum ASRError: Error, LocalizedError {
     case modelLoadFailed
     case processingFailed(String)
     case modelCompilationFailed
+    case unsupportedPlatform(String)
 
     public var errorDescription: String? {
         switch self {
@@ -86,6 +87,8 @@ public enum ASRError: Error, LocalizedError {
             return "ASR processing failed: \(message)"
         case .modelCompilationFailed:
             return "CoreML model compilation failed after recovery attempts."
+        case .unsupportedPlatform(let message):
+            return message
         }
     }
 }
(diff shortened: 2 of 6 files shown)


Sources/FluidAudioTTS/TextToSpeech/Kokoro/TtsCustomLexicon.swift
```diff
@@ -0,0 +1,287 @@
+import Foundation
+
+/// A custom pronunciation dictionary for TTS synthesis.
+///
+/// Custom lexicons allow domain-specific pronunciation overrides that take precedence
+/// over all built-in dictionaries and grapheme-to-phoneme conversion.
```
Alex-Wengg, line 6: do you think its worth adding a doc .md file instead

smdesai: Yes I believe it is, a doc as in an .md file, right? I'll take a look and provide something useful.

smdesai: Added the doc which contains python code I used for generating phonemes used for testing. I figured it'd be useful for anyone creating it and then modify it afterwards to get the desired behavior.

Alex-Wengg: i think we can remove these comments in this swift file then .

for the py code would it complex ? dependencies can be a huge issue

Sources/FluidAudioTTS/TextToSpeech/Kokoro/TtsCustomLexicon.swift
```diff
@@ -0,0 +1,287 @@
...
+/// UN=junˈITᵻd nˈAʃənz
+/// ```
+///
+/// ## Example Usage
+/// ```swift
+/// // From file
+/// let lexicon = try TtsCustomLexicon.load(from: fileURL)
+///
+/// // From string
+/// let lexicon = try TtsCustomLexicon.parse("""
```
Alex-Wengg: unit test for TtsCustomLexicon.swift ? 

Parsing valid/invalid files
Three-tier matching behavior (exact, case-insensitive, normalized)
Unicode grapheme cluster tokenization
merged(with:) conflict resolution
Edge cases: empty files, malformed lines, = in phonemes

Sources/FluidAudioCLI/Commands/TTSCommand.swift
```diff
@@ -192,7 +199,22 @@ public struct TTS {
             // Timing buckets
             let tStart = Date()
 
-            let manager = TtSManager()
+            // Load custom lexicon if provided
+            var customLexicon: TtsCustomLexicon? = nil
```
Alex-Wengg: TTSCommand.swift have the same 12-line lexicon loading block duplicated in both run() and runBenchmark():

Sources/FluidAudioTTS/TextToSpeech/Kokoro/TtsCustomLexicon.swift
smdesai: Yes my bad, it's included now.

Sources/FluidAudioCLI/Commands/TTSCommand.swift
smdesai: Refactored.

Sources/FluidAudioTTS/TextToSpeech/Kokoro/TtsCustomLexicon.swift
smdesai: Removed the comments from the swift code. The Python code is there as a reference. If someone wants to try it out, it's on them to install the dependencies.

Alex-Wengg: i am confused don't we already do these in swift ? EspeakG2P + PhonemeMapper. why do we need python misaki exactly .

smdesai: Yes they're in Swift. The Python code is just an example which I found useful. I'm happy to remove it from the doc if it causes confusion.

Alex-Wengg: i will leave it up to you but do clarify it is optional & swift supports it

Documentation/CustomPronunciationDictionary.md
```diff
@@ -0,0 +1,449 @@
...
+### Finding Valid Phonemes
+
+The Kokoro vocabulary uses a specific phoneme set. To find valid phonemes:
+
+1. Look at existing entries in the built-in lexicon
+2. Use eSpeak-NG's IPA output as a reference
+3. Test with short phrases to verify pronunciation
+
+### Phoneme Generator
+The Swift code handles the mapping of IPA tokens to Kokoro vocabulary via PhonemeMapper and EspeakG2P and the Python code here is purely an example of generating and experimenting with phonemes. It serves no other purpose. The code is based on [this](
```
Alex-Wengg: if it serves no other purposes than lets remove it

smdesai: It's removed.

Alex-Wengg:
CleanShot 2025-12-14 at 11  41 44@2x


@smdesai

Alex-Wengg (approved):

Alex-Wengg (commented):
@smdesai did you checked if xcode supports your latest change ?

smdesai:
@Alex-Wengg Yes it's tested with Xcode and runs on iOS:

https://github.com/user-attachments/assets/2cf09d8d-e7af-4c3d-8432-6c93cfd97f9a

PR #234, opened 2025-12-29
Padding blank audio to the back of short audio files

### LINKED ISSUE #232: Transcription completely blank on certain audio files
I've been setting up a streaming pipeline using Fluid, and ran into an issue. Most of the time it works great, but there are certain chunks of audio that I pass to the system, and it just completely chokes and outputs as if the audio was completely blank.

I've attached a sample project that illustrates the issue. There's a 120-second long clip (audio.mp3) that works completely fine.
While streaming this, I found that this particular 5.1-second long section within it is causing the transcriber to return a completely blank result. (51.wav). If I take the just the first 4 seconds of that 5.1-second section, it works completely fine and outputs the transcription of that section correctly.

I tried to dig into this myself, but I'm getting out of my depth. What I did find was that in `TdtDecoderV3.decodeWithTimings(...)`, there's an "inner loop" that checks for blank tokens. On the 5.1-second clip, this loop finds that every frame is blank. On the 4-second clip, it finds non-blank frames in the middle. Given that the 4-second clip is a prefix to the 5.1-second clip, I thought this was pretty unexpected.

I thought maybe someone on your team would be able to make more sense of what's going on, and tell me whether this is a bug, or whether there's something I'm doing wrong. Thanks!

[40.wav](https://github.com/user-attachments/files/24346721/40.wav)
[51.wav](https://github.com/user-attachments/files/24346723/51.wav)
[audio.mp3](https://github.com/user-attachments/files/24346722/audio.mp3)

[FluidBugTest.zip](https://github.com/user-attachments/files/24346709/FluidBugTest.zip)

Steven-Weng:
- Fix #232
- Mainly used 1 audio clip for testing because only 51.wav (in #232) was able to reproduce the blank transcript
- Adding 1 second of blank audio to the end and trimming 1 second from the beginning stops the transcription from being blank. Trimming 1 second from the beginning and adding 1 second to the end achieves the same effect which shows that the bug isn't soley based on length (could still be related to length). 
    - Adding/trimming either beginning/end also stops a blank transcript. 
    - Adding too little audio to the end still results in a blank transcript. The smallest end‑padding that still produced a blank transcript was 0.0399s. Anything larger than that stopped the blank output, which lines up with the point where the file reaches the next encoder frame.
- Fix: For short clips, I made it so that the end gets extra blank audio to reach the next encoder frame.

### DIFF
--- a/Sources/FluidAudioCLI/Commands/ASR/TranscribeCommand.swift
+++ b/Sources/FluidAudioCLI/Commands/ASR/TranscribeCommand.swift
@@ -272,7 +272,7 @@ enum TranscribeCommand {
         do {
             // Initialize ASR models
             let models = try await AsrModels.downloadAndLoad(version: modelVersion)
-            let asrManager = AsrManager()
+            let asrManager = AsrManager(config: .default)
             try await asrManager.initialize(models: models)
 
             logger.info("ASR Manager initialized successfully")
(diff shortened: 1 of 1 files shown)


Alex-Wengg:
@Steven-Weng lmk when bug is fixed then

PR #236, opened 2025-12-31
fix(diarizer): use K-Means centroids when speaker count constraint is applied

beshkenadze:
## Summary

Fixes the bug where `numSpeakers` constraint was not being applied to final diarization result.

**Root cause**: `computeCentroids()` always used `gamma/pi` from VBx output to compute centroids, ignoring the K-Means centroids stored in `vbxOutput.centroids` when `wasAdjusted=true`.

**Fix**: Check `wasAdjusted` flag first and use `vbxOutput.centroids` directly when K-Means was applied due to speaker count constraints.

## Reproduction (before fix)

```swift
let config = OfflineDiarizerConfig.default.withSpeakers(exactly: 2)
let diarizer = OfflineDiarizerManager(config: config)
let result = try await diarizer.process(audioURL)
// Expected: 2 speakers
// Actual: 3 speakers (constraint ignored)
```

## Changes

- `OfflineDiarizerManager.swift`: Added early return in `computeCentroids()` when `wasAdjusted=true`
- `SpeakerConstraintIntegrationTests.swift`: Added 5 integration tests for constraint enforcement

## Test plan

- [x] All 30 related tests pass (SpeakerConstraint, KMeans, VBxConstraint, OfflineConfig)

### DIFF
--- a/Sources/FluidAudio/Shared/AudioConverter.swift
+++ b/Sources/FluidAudio/Shared/AudioConverter.swift
@@ -1,8 +1,9 @@
-import AVFoundation
+@preconcurrency import AVFoundation
 import Accelerate
 import CoreMedia
 import Foundation
 import OSLog
+import os
 
 /// Converts audio buffers to the format required by ASR (16kHz, mono, Float32).
 ///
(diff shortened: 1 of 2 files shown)


Alex-Wengg:
hi @beshkenadze please pull the latest main branch changes to your branch. also as a headsup we have migrated to swift 6 as well

PR #247, opened 2026-01-04
fix: StreamingEouAsrManager.swift: Fix race condition causing crash a…

emrosenf:
StreamingEouAsrManager.swift: Fix race condition causing crash at removeFirst
   The process() method has an await suspension point inside a while loop.
   Between checking audioBuffer.count and calling removeFirst(), another
   actor method (like reset()) can clear the buffer, causing removeFirst()
   to crash when it tries to remove more elements than available.


Alex-Wengg (approved):

tacshi:
This fix is timely.

PR #254, opened 2026-01-12
feat(nemotron): add Nemotron Speech Streaming 0.6B support

Alex-Wengg:
## Summary

- Add streaming ASR support for NVIDIA's Nemotron Speech Streaming 0.6B model converted to CoreML
- Implement `NemotronStreamingAsrManager` with true streaming (1.12s chunks, encoder cache)
- Support int8 and float32 encoder variants (int8 is default, 4x smaller)
- Add `nemotron-benchmark` CLI command for LibriSpeech evaluation

## Performance

On LibriSpeech test-clean (100 files):
| Metric | Value |
|--------|-------|
| WER | 1.99% |
| RTFx | 8.6x |
| Memory | 1.4 GB |

## Test plan

- [x] Run `nemotron-benchmark --max-files 100` on LibriSpeech test-clean
- [x] Verify WER matches Python reference (~1.8%)
- [x] Test int8 encoder variant
- [ ] Run full LibriSpeech test-clean benchmark (2620 files)

## Usage

```bash
# Run benchmark
fluidaudiocli nemotron-benchmark --max-files 100

# With float32 encoder
fluidaudiocli nemotron-benchmark --encoder float32 --max-files 10

# With custom model directory
fluidaudiocli nemotron-benchmark --model-dir /path/to/models
```

### DIFF
--- a/Documentation/ASR/Nemotron.md
+++ b/Documentation/ASR/Nemotron.md
@@ -14,19 +14,14 @@ Nemotron Speech Streaming 0.6B is a FastConformer RNNT model optimized for strea
 
 Tested on Apple M2 with LibriSpeech test-clean:
 
-| Chunk Size | WER | RTFx | Files | Latency | Status |
-|------------|-----|------|-------|---------|--------|
-| **1120ms** | **1.99%** | 9.6x | 100 | High | Production ready |
-| **560ms** | 2.12% | 8.5x | 100 | Medium | Production ready |
-| **160ms** | ~10%* | 3.5x | 20 | Low | Limited testing |
-| **80ms** | ~60%* | 1.9x | 20 | Ultra-low | Not recommended |
+| Chunk Size | WER | RTFx | Files |
+|------------|-----|------|-------|
+| 1120ms | 1.99% | 9.6x | 100 |
+| 560ms | 2.12% | 8.5x | 100 |
+| 160ms | ~10% | 3.5x | 20 |
+| 80ms | ~60% | 1.9x | 20 |
 
-*160ms and 80ms were only tested on 20 files. Accuracy degrades significantly with smaller chunks due to insufficient context.
-
-**Notes:**
-- RTFx = Real-Time Factor (higher is better, >1x means faster than real-time)
-- 1120ms offers best accuracy; 560ms provides lower latency with minimal accuracy loss
-- 160ms/80ms are experimental - use only if ultra-low latency is critical
+160ms and 80ms were only tested on 20 files.
 
 ## Quick Start
 


Sources/FluidAudio/ASR/Streaming/NemotronStreamingAsrManager.swift
```diff
@@ -0,0 +1,528 @@
...
+public enum NemotronEncoderVariant: String, Sendable {
+    case int8 = "encoder_int8"
+    case float32 = "encoder_float32"
+
+    public var fileName: String { rawValue + ".mlmodelc" }
+}
+
+/// Configuration for Nemotron Speech Streaming 0.6B
+/// Based on nvidia/nemotron-speech-streaming-en-0.6b with 1.12s chunks
+public struct NemotronStreamingConfig: Sendable {
```
BrandonWeng, line 33: Im getting a bit worried about how many different models we have and wether it makes sesne to keep adding different managers for htem when the interface should be the same

BrandonWeng: It makes it harder to maintain this longer term and its unclear for users how and which model to use.

BrandonWeng: Im thinking if softformer is better we should just remove the older pyannote model tbh

Alex-Wengg: Not sure if there are distinct cases where pyannote dominates sorrformer

beshkenadze: I’d suggest moving model-specific configs under ASR/<ModelName> (e.g. ASR/Nemotron) so streaming managers stay generic and configs don’t get mixed. This would also make it clearer which configs belong to which model and simplify adding/removing models long-term.

Alex-Wengg: Hi @beshkenadze do you have an interest in using nemotron coreml ?

beshkenadze: I started testing FluidInference/diar-streaming-sortformer-coreml, and while I can't say for sure yet, diarization seems to work better than pyannotate.

Alex-Wengg: nemotron is nvidia's new english transcription model build for true streaming . sortformer is something else , we also have merged sortformer into main

beshkenadze: This is part of Nemotron Speech, so it is acceptable to mention it 🙂
And I plan to try nvidia/nemotron-speech-streaming-en-0.6b in captions.

Alex-Wengg: i see, we do have parakeet-eou already merged in, it functions like a smaller nemotron 0.6b hence why we were wondering if we even need to merge nemotron 0.6b when parakeet-eou already exists

Alex-Wengg:
nemotron will be put on hold possiblely 1-2 weeks from now, current priority is to merge https://github.com/FluidInference/FluidAudio/pulls

Alex-Wengg:
@beshkenadze we do not plan to implement this model given parakeet eou & parakeet tdt v3 have already operated similar  roles to nemotron

Alex-Wengg:
@nickadoodle777 and @beshkenadze nemotron is now supported in latest release

Issue #269, opened 2026-01-29
v0.10.1 SortformerModelInference.swift code needs architecture checks


schmatz:
Without checking for the arm64 architecture, you get build errors in the Xcode GUI, even if you explicitly exclude Intel builds in a variety of ways. Claude writeup below.

Thanks for this library by the way, it's extremely useful!

# Float16 Availability Issue in SortformerModelInference.swift

## Summary

FluidAudio 0.10.1 fails to build when archiving for macOS because `Float16` is used without an architecture guard. The `Float16` type is only available on Apple Silicon (arm64), not Intel (x86_64), regardless of macOS version.

## Environment

- Xcode 16.2
- macOS 15.6+ deployment target
- FluidAudio 0.10.1 (commit `0afbabca218c6c0e363fc128e1b50be4e69abc5e`)

## Problem

When archiving a macOS app that depends on FluidAudio, the build fails with:

```
SortformerModelInference.swift:254:80 Conformance of 'Float16' to 'MLShapedArrayScalar' is unavailable in macOS
SortformerModelInference.swift:254:101 'Float16' is unavailable in macOS
SortformerModelInference.swift:255:18 Conformance of 'Float16' to 'MLShapedArrayScalar' is unavailable in macOS
SortformerModelInference.swift:257:42 No exact matches in call to initializer
```

**Root cause**: Xcode Archive builds for distribution, which by default includes both arm64 and x86_64 architectures. When compiling for x86_64, the `Float16` type is not available, causing the build to fail.

The existing `#available(macOS 15.0, iOS 18.0, *)` check only guards against OS version, not CPU architecture. `Float16` requires **both** macOS 15+ **and** arm64 architecture.

## Location

`Sources/FluidAudio/Diarizer/Sortformer/SortformerModelInference.swift`, lines 253-260

## Current Code (Broken)

```swift
// Chunk embeddings may be Float16 (head module uses fp16) or Float32
let chunkEmbeddings: [Float]
if let fp32 = output.featureValue(for: "chunk_pre_encoder_embs_out")?.shapedArrayValue(of: Float32.self)?
    .scalars
{
    chunkEmbeddings = fp32
} else if #available(macOS 15.0, iOS 18.0, *),
    let fp16 = output.featureValue(for: "chunk_pre_encoder_embs_out")?.shapedArrayValue(of: Float16.self)?
        .scalars
{
    chunkEmbeddings = fp16.map { Float($0) }
} else {
[...]

PR #270, opened 2026-01-29
Fix Float16 build failure in Xcode GUI by adding architecture checks

schmatz:
See #269 for motivation and details.

Add #if arch(arm64) guard around Float16 usage in SortformerModelInference.swift.

Float16 is only available on Apple Silicon, not Intel, regardless of macOS version. The existing #available check only guards OS version, not CPU architecture. This causes build failures when archiving macOS apps for distribution (which builds for both arm64 and x86_64 by default). I tried to work around this with every setting I could find but did not get it working in the GUI (CLI worked.)

Fixes build error:
- 'Float16' is unavailable in macOS
- Conformance of 'Float16' to 'MLShapedArrayScalar' is unavailable in macOS

Allows building Mac apps that use this with the Xcode GUI


Alex-Wengg:
@schmatz its gonna be a while until we make a release tag as we are cleaning some things up . if that is fine i will merge this pr asap

PR #326, opened 2026-02-20
fix: convert AsrManager to actor for Swift 6 concurrency safety

### LINKED ISSUE #322: Swift 6 strict concurrency errors in StreamingAsrManager: "Sending 'asrManager' risks causing data races"
**Summary**
When consuming FluidAudio from a project with Swift 6 strict concurrency checking enabled (SWIFT​_​STRICT​_​CONCURRENCY​=complete), the build fails with errors in Streaming​Asr​Manager​.swift:

```
error: Sending 'asrManager' risks causing data races
```

This occurs at lines 258, 379, and 621.

**Environment**
• FluidAudio: current main (commit 99220bc4)
• Swift: 6.0
• Xcode: 16.3
• Consumer project settings: SWIFT​_​STRICT​_​CONCURRENCY​=complete, SWIFT​_​TREAT​_​WARNINGS​_​AS​_​ERRORS​=​YES

**Problem**
Streaming​Asr​Manager is an actor that stores:
```
nonisolated(unsafe) private var asrManager: AsrManager?
```

The code pattern causing the error:

```
guard let asrManager = asrManager else { return }
try await asrManager.resetDecoderState(for: audioSource)  // Error here
```

The local let asr​Manager binding captures the non-Sendable Asr​Manager reference. When this captured reference crosses the await boundary, Swift 6 strict concurrency checking reports "Sending 'asrManager' risks causing data races" because Asr​Manager is a non-Sendable class.

**Why This Happens**
Even though:
1. Streaming​Asr​Manager is an actor (providing isolation)
2. The property is marked nonisolated(unsafe) (acknowledging manual safety)

...the local binding let asr​Manager = asr​Manager creates a separate captured reference. Swift sees this captured value potentially being accessed from another context after the await suspension point.

**Proposed Fix**

Mark Asr​Manager as @unchecked ​Sendable:

```
// In AsrManager.swift, line 11
public final class AsrManager: @unchecked Sendable {
```

**Rationale:**
1. Asr​Manager is already used across async boundaries throughout the codebase
2. The existing comment in Streaming​Asr​Manager​.swift (lines 20-21) acknowledges manual safety: "AsrManager contains CoreML models which are not Sendable. We manage the safety ourselves by only accessing it from within the actor."
3. This pattern is consistent with how the codebase already handles Ctc​Keyword​Spotter and Vocabulary​Rescorer (also marked nonisolated(unsafe))

**Alternative Fix**
If @unchecked ​Sendable is not preferred, the actor methods could avoid creating local bindings:
```
// Instead of:
guard let asrManager = asrManager else { return }
try await asrManager.resetDecoderState(for: audioSource)

// Use:
guard let manager = self.asrManager else { return }
try await manager.resetDecoderState(for: audioSource)
```

However, this is functionally identical and likely won't resolve the issue - the @unchecked ​Sendable conformance is the correct fix.

**Affected Lines**
• Streaming​Asr​Manager​.swift​:258 - in reset() method
• Streaming​Asr​Manager​.swift​:379 - in process​Streaming​Chunk() method
• Streaming​Asr​Manager​.swift​:621 - in reset​Decoder​For​Recovery() method

claudeaceae:
## Summary

Fixes #322 — resolves strict concurrency errors in `StreamingAsrManager` by converting `AsrManager` from a `class` to an `actor`, eliminating the need for `nonisolated(unsafe)` or `@unchecked Sendable`.

### Changes

- **AsrManager.swift**: `public final class` → `public actor`; replaced `lazy var` with initialized `var`; internal `inout` decoder state mutations use local-copy-and-assign pattern to satisfy actor isolation rules
- **AsrTranscription.swift**: Added `executeMLInferenceReturningState` wrapper for cross-actor callers (returns updated state instead of `inout`); marked `padAudioIfNeeded` as `nonisolated` (pure function with no state access)
- **ChunkProcessor.swift**: Updated `transcribeChunk` from `inout TdtDecoderState` to pass-and-return pattern; uses new `executeMLInferenceReturningState`; added `await` to `processTranscriptionResult` calls
- **StreamingAsrManager.swift**: Removed `nonisolated(unsafe)` from `asrManager`, `ctcSpotter`, and `vocabularyRescorer` properties; added `await` at cross-actor call sites; simplified optional handling in `reset()` and `resetDecoderForRecovery()`
- **TranscribeCommand.swift / TTSCommand.swift**: Added `await` to `cleanup()` calls

### Approach

Per @BrandonWeng's [comment](https://github.com/FluidInference/FluidAudio/issues/322#issuecomment-2798744505), this uses the actor-based approach rather than `@unchecked Sendable`. The key challenge was that `inout` parameters can't cross actor boundaries, so:

1. **Actor-internal mutations** (e.g. `transcribe()`, `resetDecoderState()`) use a copy-to-local → mutate → assign-back pattern
2. **Cross-actor calls** (from `ChunkProcessor`) use a new `executeMLInferenceReturningState` method that wraps the `inout` version and returns the updated state as a tuple element

## Test plan

- [x] `swift build` passes with no errors or warnings
- [x] `swift-format` applied to all modified files
- [ ] Run full test suite (CI)
- [ ] Verify streaming transcription works end-to-end on device

### DIFF
--- a/Sources/FluidAudio/TTS/PocketTTS/Assets/PocketTtsResourceDownloader.swift
+++ b/Sources/FluidAudio/TTS/PocketTTS/Assets/PocketTtsResourceDownloader.swift
@@ -8,9 +8,12 @@ public enum PocketTtsResourceDownloader {
     private static let logger = AppLogger(category: "PocketTtsResourceDownloader")
 
     /// Ensure all PocketTTS models are downloaded and return the cache directory.
-    public static func ensureModels() async throws -> URL {
-        let cacheDirectory = try cacheDirectory()
-        let modelsDirectory = cacheDirectory.appendingPathComponent(
+    ///
+    /// - Parameter directory: Optional override for the base cache directory.
+    ///   When `nil`, uses the default platform cache location.
+    public static func ensureModels(directory: URL? = nil) async throws -> URL {
+        let targetDir = try directory ?? cacheDirectory()
+        let modelsDirectory = targetDir.appendingPathComponent(
             PocketTtsConstants.defaultModelsSubdirectory)
 
         let repoDir = modelsDirectory.appendingPathComponent(Repo.pocketTts.folderName)
@@ -36,8 +39,10 @@ public enum PocketTtsResourceDownloader {
     ///
     /// This is an optional model that's only needed for voice cloning functionality.
     /// It's downloaded separately from the main models to reduce initial download size.
-    public static func ensureMimiEncoder() async throws -> URL {
-        let repoDir = try await ensureModels()
+    /// - Parameter directory: Optional override for the base cache directory.
+    ///   When `nil`, uses the default platform cache location.
+    public static func ensureMimiEncoder(directory: URL? = nil) async throws -> URL {
+        let repoDir = try await ensureModels(directory: directory)
         let encoderPath = repoDir.appendingPathComponent(ModelNames.PocketTTS.mimiEncoderFile)
 
         if FileManager.default.fileExists(atPath: encoderPath.path) {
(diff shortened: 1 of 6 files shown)


Alex-Wengg:
@BrandonWeng this looks ready for review

Steven-Weng:
See my comment on the issue: https://github.com/FluidInference/FluidAudio/issues/322#issuecomment-3940191517

PR #118, opened 2025-09-24
Use independent Melspec and Encoder models for iPhones 

BrandonWeng:
### Why is this change needed?
Compile time Metrics on iPhone 16 Pro Max:

Before, 20,000ms (5mins+) to compile encoder and processor

```
Updating content for activity 34DE0CCF-64A6-455F-BE25-20AFCB9CFDC4
Updating content for activity 34DE0CCF-64A6-455F-BE25-20AFCB9CFDC4
Loaded model: MelEncoder.mlmodelc
Compiled ASR model MelEncoder.mlmodelc in 20018.84 ms
Updating content for activity 34DE0CCF-64A6-455F-BE25-20AFCB9CFDC4
Loaded model: Decoder.mlmodelc
Compiled ASR model Decoder.mlmodelc in 116.61 ms
Loaded model: JointDecision.mlmodelc
Compiled ASR model JointDecision.mlmodelc in 122.83 ms
Successfully downloaded ASR models
```

After: 
```
Loaded model: Encoder.mlmodelc
Compiled ASR model Encoder.mlmodelc in 3361.23 ms
Found parakeet-tdt-0.6b-v3-coreml locally, no download needed
Updating content for activity 18AB8F87-3186-4EE1-BE7D-C2F5893E29E6
Loaded model: Decoder.mlmodelc
Compiled ASR model Decoder.mlmodelc in 88.49 ms
Found parakeet-tdt-0.6b-v3-coreml locally, no download needed
Loaded model: JointDecision.mlmodelc
Compiled ASR model JointDecision.mlmodelc in 48.46 ms
Successfully downloaded ASR models
Loading ASR models from: /var/mobile/Containers/Data/Application/94B9177E-C7FC-465C-82E9-7BF21F0F971A/Library/Application Support/FluidAudio/Models/parakeet-tdt-0.6b-v3-coreml
Found parakeet-tdt-0.6b-v3-coreml locally, no download needed
Loaded model: Preprocessor.mlmodelc
Compiled ASR model Preprocessor.mlmodelc in 9.15 ms
Loaded Preprocessor.mlmodelc with compute units: MLComputeUnits(rawValue: 2)
Found parakeet-tdt-0.6b-v3-coreml locally, no download needed
Loaded model: Encoder.mlmodelc
Compiled ASR model Encoder.mlmodelc in 162.05 ms
Loaded Encoder.mlmodelc with compute units: MLComputeUnits(rawValue: 1)
[...]

PR #383, opened 2026-03-16
feat: support parakeet-tdt-ctc-110m hybrid model

JarbasAl:
Add AsrModelVersion.tdtCtc110m for the 110M parameter hybrid TDT-CTC model. Key differences from the 0.6B models:

- Fused preprocessor+encoder (no separate Encoder.mlmodelc)
- Smaller dimensions: encoderHidden=512, vocabSize=1024, 1 LSTM layer
- Array-format vocabulary (vocab.json) instead of dict format
- blankId=1024 (same as v2)

Changes:
- AsrModels: optional encoder, fused frontend loading, array vocab support
- AsrManager: version-aware decoder state shapes, fused frontend availability
- AsrTranscription: skip encoder step when preprocessor output is fused
- TdtDecoderState: parameterized LSTM layer count
- TdtDecoderV3: use config.encoderHiddenSize instead of auto-detection
- EncoderFrameView: accept explicit hidden size parameter
- TranscribeCommand: --model-version tdt-ctc-110m, --model-dir flags
- ModelNames: parakeetTdtCtc110m repo, fused model requirements

Companion PR: https://github.com/FluidInference/mobius/pull/25

better support for https://huggingface.co/nvidia/parakeet-tdt_ctc-110m

### AI Disclosure

I never worked with swift before, Claude Opus did most of the work

---

<a href="https://app.devin.ai/review/fluidinference/fluidaudio/pull/383" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/Sources/FluidAudio/ASR/AsrManager.swift
+++ b/Sources/FluidAudio/ASR/AsrManager.swift
@@ -290,6 +290,7 @@ public final class AsrManager {
 
     public func cleanup() {
         let layers = asrModels?.version.decoderLayers ?? 2
+        asrModels = nil
         preprocessorModel = nil
         encoderModel = nil
         decoderModel = nil


Alex-Wengg:
@JarbasAl did you test this on iOS , we had originally fused preprocessor+encoder before & it had incompatibility issues on iOS .

also what about the benchmarks

Sources/FluidAudio/ModelNames.swift
```diff
@@ -16,6 +16,7 @@ public enum Repo: String, CaseIterable {
     case qwen3Asr = "FluidInference/qwen3-asr-0.6b-coreml/f32"
     case qwen3AsrInt8 = "FluidInference/qwen3-asr-0.6b-coreml/int8"
     case multilingualG2p = "FluidInference/charsiu-g2p-byt5-coreml"
+    case parakeetTdtCtc110m = "FluidInference/parakeet-tdt-ctc-110m-coreml"
```
Alex-Wengg, line 19: we don't have this on FluidInference HF

JarbasAl: I assumed you would upload it before merging, also sent companion PR to mobius for the conversion

JarbasAl:
> @JarbasAl did you test this on iOS , we had originally fused preprocessor+encoder before & it had incompatibility issues on iOS .
> 
> also what about the benchmarks

I only tested in a Mac mini, not iOS.  but I should note I had to use iOS18 target for the conversion to work

EDIT: I take that back, works with 17

Alex-Wengg:
hi @JarbasAl 
thanks for the contribution! what's the intended use case for this model? The differences listed in the description (fused frontend, smaller hidden size, array vocab) are structural traits rather than advantages over the 0.6B. what motivated of the conversion of this model?

JarbasAl:
> hi @JarbasAl thanks for the contribution! what's the intended use case for this model? The differences listed in the description (fused frontend, smaller hidden size, array vocab) are structural traits rather than advantages over the 0.6B. what motivated of the conversion of this model?

I am developing an application with FluidAudio where I use a proprietary finetuned version of that model, STT is the odd component not using FluidAudio directly.

figured it could be useful for the community to share support, the 100m model is very lightweight

WER improved ~3% in my test data by using this instead of the CTC export

Sources/FluidAudio/ASR/TDT/TdtDecoderState.swift
```diff
@@ -24,15 +24,15 @@ struct TdtDecoderState: Sendable {
     /// - zero: Decoder exactly at the end of encoder frames
     var timeJump: Int?
 
-    init() throws {
+    init(decoderLayers: Int = 2) throws {
```
Alex-Wengg, line 27: any reasons why int = 2?

Sources/FluidAudio/ASR/ChunkProcessor.swift
```diff
@@ -65,7 +65,9 @@ struct ChunkProcessor {
 
         var chunkStart = 0
         var chunkIndex = 0
-        var chunkDecoderState = TdtDecoderState.make()
+        var chunkDecoderState = TdtDecoderState.make(
+            decoderLayers: manager.asrModels?.version.decoderLayers ?? 2
```
Alex-Wengg, line 69: 2 ?

Alex-Wengg:
https://huggingface.co/FluidInference/parakeet-tdt-ctc-110m-coreml HF uploaded

Alex-Wengg:
> We probably need to double check if this runs on iOS or not. We previously had issues with ioS when we tried combining the mel processor and the encoder.
> 
> #118
> 
> If there's no problem, @Alex-Wengg can't we just replace the existing ctc 110m with this instead of maintaining both

this is in theory possible but i will need to do some testings first. the custom vocab research paper did not  mention anything about preprocessor specifications and generally preprocessors are pretty simple too

Alex-Wengg:
verified iOS deployment and inference  https://github.com/FluidInference/FluidAudio/pull/433 
image

Alex-Wengg:
worked upon & resolved in https://github.com/FluidInference/FluidAudio/pull/433

PR #422, opened 2026-03-25
fix: computeFlat mel frame count off-by-one for center-padded audio

makeusabrew:
Center padding adds nFFT/2 (256 samples) on each side before windowing. The leading pad produces one additional STFT frame that the original formula (audioCount / hopLength) misses. This causes the EOU streaming model to receive 63 mel frames instead of the expected 64, triggering a CoreML shape mismatch error on every inference.

Fix: 1 + audioCount / hopLength (matches the center-padded frame count). Only affects computeFlat — computeFlatTransposed is used by Sortformer which handles the frame count differently via melLength * melStride.

---

<a href="https://app.devin.ai/review/fluidinference/fluidaudio/pull/422" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/Sources/FluidAudio/Shared/AudioMelSpectrogram.swift
+++ b/Sources/FluidAudio/Shared/AudioMelSpectrogram.swift
@@ -190,7 +190,10 @@ public final class AudioMelSpectrogram {
         C.Element == Float, C.Index == Int
     {
         let audioCount = audio.count
-        let numFrames = audioCount / hopLength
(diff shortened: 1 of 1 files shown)


makeusabrew:
Apologies, opened in error. We *are* seeing issues with 0.12.6 using the `StreamingEouAsrManager`, but I'm not yet convinced this patch is correct.

PR #376, opened 2026-03-15
LS-EEND Diarizer

SGD2718:
---

## Add LS-EEND speaker diarization

Sortformer handles up to 4 speakers and works best at 16 kHz in noisy environments. That leaves a gap for phone calls, large meetings, and recordings with unknown conditions. LS-EEND fills it: up to 10 speakers (variant-dependent), trained on telephone, meeting, and in-the-wild corpora, operating at 8 kHz.

This PR adds LS-EEND as a first-class diarizer alongside Sortformer — same `Diarizer` protocol, same CLI patterns, same post-processing pipeline.

### Why these changes are needed

**Unified timeline** — `SortformerTimeline` was Sortformer-specific and couldn't be shared. LS-EEND needs the same post-processing (threshold, median filter, onset/offset padding, min-duration filtering, finalized vs tentative segments). `DiarizerTimeline` replaces `SortformerTimeline` with a shared implementation that both models use, eliminating duplicated logic.

**LS-EEND diarizer** — The model was partially wired up but missing a clean public API, proper `Diarizer` protocol conformance, and integration with `DiarizerTimeline`. This completes the implementation: offline file processing with automatic resampling, streaming with committed + speculative preview frames, and session-level control via `LSEENDStreamingSession`.

**CLI** — Without `lseend` and `lseend-benchmark`, the model can't be used or evaluated outside of Swift code. The benchmark also validates that DER matches the paper's reported numbers before shipping to users.

**AMI ground truth fallback** — `lseend-benchmark --variant ami` silently produced no results because the benchmark looked for RTTM files that don't exist in the standard dataset layout. Added the same `AMIParser` XML annotation fallback that the Sortformer benchmark uses.

**Tests** — `LSEENDRuntimeTests` runs the inference engine, streaming session, and feature extractor against known-good outputs to catch regressions in the CoreML pipeline.

**Documentation** — LS-EEND has a substantially different API surface than Sortformer (five source files, streaming session layer, matrix type, full evaluation namespace, per-variant speaker caps). Documents the entire public API and provides a variant selection guide.

### Changes

**`DiarizerTimeline.swift`** (new) — Unified post-processing timeline shared by both Sortformer and LS-EEND. Replaces `SortformerTimeline.swift` (deleted). `SortformerDiarizerPipeline` updated to use it.

[...]

PR #444, opened 2026-03-27
Fix EOU frame count calculation for center-padded mel spectrograms

### LINKED ISSUE #441: Creating StreamingEouAsrManager with 320ms is broken
HI - from v0.12.5+ setting the chunk size to 320ms breaks. The console log generates:

```
EOU process error: MultiArray shape (1 x 128 x 63) does not match the shape (1 x 128 x 64) specified in the model description
multiArrayConstraint isAllowedValue: fails for MLFeatureTypeMultiArray feature audio_signal with featureValue MultiArray : Float32 1 × 128 × 63 array. Error description: MultiArray
```

So the shape is wrong...

- Root Cause Analysis

The root cause is the current AudioMelSpectrogram.computeFlat uses : 

`numFrames = audioCount / hopLength (= 63 for ms320)
`

but the old NeMoMelSpectrogram.computeFlat it replaced used :

` numFrames = 1 + (paddedCount - winLength) / hopLength (= 64 for ms320). 
`

The padding formula is correct because computeFlat pads the audio by nFFT/2 on each side before processing.

- (Possible) Solution

Claude has proposed this fix - I tested it and it works.

```
    public func computeFlat<C>(
        audio: C, lastAudioSample: Float = 0
    ) -> (mel: [Float], melLength: Int, numFrames: Int)
    where
        C: AccelerateBuffer & Collection,
        C.Element == Float, C.Index == Int
    {
        let audioCount = audio.count
        // Frame count matches NeMo's center-padded mel: audio is zero-padded by nFFT/2 on each side
        // before windowing, so numFrames = 1 + (paddedCount - winLength) / hopLength.
        let padLength = nFFT / 2
        let paddedCount = audioCount + 2 * padLength
        let numFrames = 1 + (paddedCount - winLength) / hopLength

        guard numFrames > 0, let firstSample = audio.first else {
            return (mel: [Float](repeating: padValue, count: nMels), melLength: 0, numFrames: 1)
        }

        // Ceil(numFrames / padTo) * padTo
        let numPaddedFrames: Int = ((numFrames - 1) / padTo + 1) * padTo

        // Step 1: Apply preemphasis filter using vDSP (y[n] = x[n] - preemph * x[n-1])
        // This will be copied into an already padded buffer to save time.

      ....
      ....
}
```

- My Environment
-- MacBookPro - M1 Pro - 16GB
-- macOS 26.4
-- Xcode 26.4
-- FluidAudio (local pinned at version v0.12.6 + swift-transformers 1.1.6

Alex-Wengg:
## Summary

Fixes #441 - StreamingEouAsrManager with 320ms chunks was producing incorrect frame counts, causing shape mismatches.

- Updated `AudioMelSpectrogram.computeFlat()` to use correct frame count formula
- Updated `AudioMelSpectrogram.computeFlatTransposed()` with `.center` padding mode
- Changed from `numFrames = audioCount / hopLength` to `numFrames = 1 + (paddedCount - winLength) / hopLength`
- This accounts for nFFT/2 center padding applied before STFT processing, matching NeMo's computation

## Root Cause

The original formula didn't account for the center padding (nFFT/2 on each side) that's applied to audio before windowing. This caused the frame count to be off by 1, producing 63 frames instead of 64 for 630ms audio chunks.

### Frame Count Validation Tests
Added `EouChunkSizeFrameCountTests` - all passing:
- ✅ 160ms: 17 frames (was 16)
- ✅ 320ms: 64 frames (was 63) ← **Issue #441 error case**
- ✅ 1280ms: 129 frames (was 128)
- ✅ Tested with 10 different audio lengths per chunk size

### Integration Tests (10 files per chunk size)
**30 transcriptions total - 100% success rate:**

| Chunk Size | Files | Success | Avg WER | Overall WER |
|------------|-------|---------|---------|-------------|
| 160ms | 10/10 | 100% | 8.40% | 9.64% |
| 320ms | 10/10 | 100% | 4.92% | 5.72% |
| 1280ms | 10/10 | 100% | 7.19% | 7.83% |

**✅ No shape mismatch errors detected across all 30 transcriptions**

The 320ms chunk size (the problematic one from issue #441) now works perfectly and actually achieves the lowest WER!

## Test Plan

- [x] All `AudioMelSpectrogramTests` pass
- [x] Added `EouChunkSizeFrameCountTests` - all passing
- [x] Integration test: 10 files × 3 chunk sizes = 30 successful transcriptions
- [x] WER calculation confirms transcription quality maintained (5-10% WER)
- [x] Verified no shape mismatch errors

All tests pass successfully.

### DIFF
--- a/Sources/FluidAudio/DownloadUtils.swift
+++ b/Sources/FluidAudio/DownloadUtils.swift
@@ -67,7 +67,7 @@ public class DownloadUtils {
             case .downloadFailed(let path, let underlying):
                 return "Failed to download \(path): \(underlying.localizedDescription)"
             case .htmlErrorResponse(let path, let snippet):
-                return "HuggingFace returned HTML error page instead of JSON for \(path). This usually indicates rate limiting or server issues. Response snippet: \(snippet)"
+                return "HuggingFace returned HTML instead of JSON for \(path) (rate limit or server issue): \(snippet)"
             case .modelNotFound(let path):
                 return "Model file not found: \(path)"
             }


Alex-Wengg:
> We independently traced this same bug from the Zoryn (GlibScribe) side — the 160ms EOU streaming path was hitting the shape mismatch on every audio chunk, causing a full fallback to Apple Speech for L1 streaming.
> 
> **Root cause confirmed**: PR #376 (LS-EEND) accidentally replaced `computeFlat()`'s correct center-padded formula (`1 + (paddedCount - winLength) / hopLength` → 17 frames) with the Sortformer-specific formula from `computeFlatTransposed()` (`audioCount / hopLength` → 16 frames) during the move to `Shared/`.
> 
> We've cherry-picked `05a59f3` onto our fork (`VillaAI/audio-engine` branch `fix/eou-mel-frame-count`) and are testing it in Zoryn now. Would be great to get this merged to main so downstream consumers pick it up.

this will be merged today.

PR #543, opened 2026-04-26
feat(tts/kokoro): expose per-phoneme and per-word timing on ChunkInfo

Zoxso:

The Kokoro Core ML model already returns a `pred_dur` feature on every prediction — one frame count per input phoneme — but `KokoroSynthesizer` currently sums it for total-length cropping at [`KokoroSynthesizer.swift:417`](https://github.com/FluidInference/FluidAudio/blob/main/Sources/FluidAudio/TTS/Kokoro/Pipeline/Synthesize/KokoroSynthesizer.swift#L417) and discards the per-token detail. That makes downstream features that need per-word timing (read-along / karaoke highlighting, lyrics-style captions, time-aligned subtitles, forced-alignment-style use cases) impractical without bolting on a second model (e.g. `SFSpeechRecognizer` or a CTC aligner) just to recover information the TTS model already produced.

This PR surfaces that information without changing any existing behavior.

### What changed

Two new optional public fields on `KokoroSynthesizer.ChunkInfo`:

```swift
public let tokenTimings: [TokenTiming]?   // one entry per input phoneme
public let wordTimings:  [WordTiming]?    // aggregated to atom granularity
```

```swift
public struct TokenTiming: Sendable {
    public let phoneme: String
    public let startTime: TimeInterval     // seconds, from chunk start
    public let endTime:   TimeInterval
    public let frames:    Float            // raw pred_dur value
}

public struct WordTiming: Sendable {
    public let word:       String          // matches an entry in ChunkInfo.atoms
    public let atomIndex:  Int
    public let startTime:  TimeInterval
    public let endTime:    TimeInterval
}
```

Both default to `nil` so existing callers and existing `ChunkInfo` initializer sites compile unchanged. When the underlying Core ML model exposes `pred_dur` (the current behavior on v0.14.1), both fields are populated for every successful chunk. The `voiceSpeed`-adjusted pass scales timings by `1/factor` to line up with the time-stretched samples.

### How it works

1. `synthesizeChunk` now returns `(samples, predictionTime, predDur)`. The existing `pred_dur` extraction stays put; the array is sliced to the true input length (padding excluded) and threaded through `ChunkSynthesisResult`.
2. The internal `TextChunk` (chunker output) gains `phonemeRanges: [Range<Int>]` — for each `atoms[i]`, the range of `phonemes[]` indices it produced. Punctuation atoms get a one-phoneme range; inter-word separator phonemes (single-space tokens) are intentionally not attributed to any atom.
3. Two new static helpers convert raw `pred_dur` values into the public timing types using `TtsConstants.kokoroFrameSamples` (600) and `TtsConstants.audioSampleRate` (24_000) — i.e. each frame = 25 ms.

### Tests

- `KokoroTokenTimingTests` (11 cases) — covers the two helpers' math, monotonicity, length-mismatch returning nil, partial-range clipping, and out-of-bounds atom skipping. No model required.
- `KokoroChunkerPhonemeRangesTests` (3 cases) — verifies the chunker produces correct phoneme ranges across word, punctuation, and multi-word inputs, and that ranges are non-overlapping.

Existing test suite (1526 tests) passes unchanged on macOS arm64; 24 environment-skipped tests are unaffected by this change.

### Source compatibility

Fully additive — every existing call to `ChunkInfo.init` still compiles because the two new parameters default to `nil`. No behavior change for callers that don't read the new fields.

### Out of scope

- Multi-language phoneme alignment (this PR uses the chunker's existing US-English alignment surface, matching the current Beta TTS scope)
- Cross-chunk timing reconciliation (consumers already know `samples.count` per chunk)
- Forced-alignment refinement (raw model output is the source of truth)
- Any changes to model files, weights, or vocabularies

---

<a href="https://app.devin.ai/review/fluidinference/fluidaudio/pull/543" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/Sources/FluidAudio/TTS/Kokoro/Pipeline/Postprocess/KokoroSynthesizer+Types.swift
+++ b/Sources/FluidAudio/TTS/Kokoro/Pipeline/Postprocess/KokoroSynthesizer+Types.swift
@@ -45,6 +45,47 @@ extension KokoroSynthesizer {
         }
     }
 
+    /// Predicted duration for a single phoneme/input token, derived from the model's `pred_dur` output.
+    /// Times are in seconds, measured from the start of the chunk's audio (`ChunkInfo.samples`).
+    public struct TokenTiming: Sendable {
+        /// Phoneme glyph from the chunker's vocabulary lookup.
+        public let phoneme: String
+        /// Cumulative start time in seconds, measured from the start of this chunk's audio.
+        public let startTime: TimeInterval
+        /// End time in seconds, measured from the start of this chunk's audio.
+        public let endTime: TimeInterval
+        /// Raw frame count from the model's `pred_dur` output.
+        /// One frame = `TtsConstants.kokoroFrameSamples` (600) samples = 25 ms at 24 kHz.
+        public let frames: Float
+
+        public init(phoneme: String, startTime: TimeInterval, endTime: TimeInterval, frames: Float) {
+            self.phoneme = phoneme
+            self.startTime = startTime
+            self.endTime = endTime
+            self.frames = frames
+        }
+    }
+
+    /// Word-level timing aggregated from `TokenTiming`s using the chunker's phoneme→atom alignment.
+    /// Times are in seconds, measured from the start of the chunk's audio (`ChunkInfo.samples`).
+    public struct WordTiming: Sendable {
+        /// Source word/atom text — matches an entry in `ChunkInfo.atoms`.
+        public let word: String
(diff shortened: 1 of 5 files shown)


Alex-Wengg:
please address the comments left by Devin

Alex-Wengg:
also is there a way to test this change accurately do we have a benchmark dataset ?

Alex-Wengg:
closing due to lack of activity & response

PR #568, opened 2026-05-03
Finalized DiarzerTimeline segment updates no longer commit tentative segments

SGD2718:
There was a bug that would cause the trailing diarizer segment to disappear if minFramesOff was nonzero once the person stopped talking.


Alex-Wengg:
@SGD2718 i am gonna do a release soon later for a bunch of other changes

PR #591, opened 2026-05-09
Make SpeakerManager a struct and de-async DiarizerManager

panv-kw:

`DiarizerManger.performCompleteDiarization` is `async`, even though no asynchronous operations occur when running the models and processing the results - this is just plain, synchronous computation. It doesn't wait on the network or things like that. It is important to be able to integrate it in to other synchronous compute workflows.

The reason it had to be `async` until now is that the `SpeakerManager` type containing the speaker database was a `class`, meaning that it was shared mutable state. It was made an `actor` because this shared mutable state could be mutated concurrently.

But really, there should not be concurrent mutations to the `SpeakerManager` in the first place. The user of this type, `DiarizationManager`, is not actually prepared for other code to be modifying this database while it is using it, and anybody who is trying is almost certainly writing a bug because their code would be logically racing with `DiarizationManager` and the results would be unpredictable.

This change makes `SpeakerManager` a struct. It has copy-on-write value semantics because it wraps a `Dictionary` for its storage, and mutations are marked by the `mutating` keyword and require exclusive ownership of the variable -- again, just like `Dictionary`. The compiler statically diagnoses attempts to concurrently mutate the `DiarizationManager`'s speaker database, so the test for this can be removed (it no longer compiles).

Screenshot 2026-05-09 at 19 10 26

In summary, this change significantly reduces the cognitive load of using and maintaining this code, promotes correct usage through static diagnostics rather than allowing unpredictable results through concurrent mutation of the speaker database, and enables diarization to be used in more contexts in more programs.

(BTW, `SpeakerManager` doesn't strictly _need_ to be `Sendable`, but the previous one was by virtue of being an `actor`, so I marked this one as being `Sendable` too in case anybody was relying on it. I don't think the implementation of this type is going to change radically in the future to the point where that might be a problem)

### DIFF
--- a/Sources/FluidAudio/Diarizer/Clustering/SpeakerManager.swift
+++ b/Sources/FluidAudio/Diarizer/Clustering/SpeakerManager.swift
@@ -5,12 +5,7 @@ import OSLog
 /// In-memory speaker database for streaming diarization
 /// Tracks speakers across chunks and maintains consistent IDs
 ///
-/// This is an `actor` to provide thread-safe access under Swift 6 strict
-/// concurrency. Previous implementations used a `DispatchQueue` which could
-/// trigger `unsafeForcedSync` warnings (and occasional libmalloc heap
-/// corruption via concurrent `[Float]` COW mutations) when called from
-/// async contexts such as VAD / diarization buffer callbacks.
-public actor SpeakerManager {
+public struct SpeakerManager: Sendable {
     internal let logger = AppLogger(category: "SpeakerManager")
 
     // Constants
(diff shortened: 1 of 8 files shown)


panv-kw:
Another example of static concurrency diagnostics:

Add the following to `Sources/FluidAudioCLI/Commands/DiarizationBenchmark.swift`, at line 445:

```swift
            // Configure streaming manager
            diarizerManager.speakerManager.speakerThreshold = assignmentThreshold
            diarizerManager.speakerManager.embeddingThreshold = updateThreshold

            Task { diarizerManager.speakerManager.assignSpeaker([], speechDuration: 3.3) }  // 👈

            // Process in chunks
            let samplesPerChunk = Int(chunkSeconds * 16000)
            let hopSamples = Int((chunkSeconds - overlapSeconds) * 16000)
```

And you will see that the compiler doesn't allow it, and complains that all the other code using the `diarizerManager` would race with this:

```
FluidAudio/Sources/FluidAudioCLI/Commands/DiarizationBenchmark.swift:445:13: error: sending value of non-Sendable type '() async -> Speaker?' risks causing data races
            Task { diarizerManager.speakerManager.assignSpeaker([], speechDuration: 3.3) }
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
FluidAudio/Sources/FluidAudioCLI/Commands/DiarizationBenchmark.swift:445:13: note: Passing value of non-Sendable type '() async -> Speaker?' as a 'sending' argument to initializer 'init(name:priority:operation:)' risks causing races in between local and caller code
            Task { diarizerManager.speakerManager.assignSpeaker([], speechDuration: 3.3) }
            ^
FluidAudio/Sources/FluidAudioCLI/Commands/DiarizationBenchmark.swift:571:53: note: access can happen concurrently
            let finalSpeakerCount = diarizerManager.speakerManager.speakerCount
                                    ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
FluidAudio/Sources/FluidAudioCLI/Commands/DiarizationBenchmark.swift:478:55: note: access can happen concurrently
                let chunkResult = try diarizerManager.performCompleteDiarization(
                                      ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

So concurrent mutation is blocked statically.

Alex-Wengg:
@panv-kw i haven't seen you for a while how you been.

PR #630, opened 2026-05-20
fix(asr): reduce English drift on French recordings via token blocklist

Matth-93:
Parakeet TDT v3 has no language conditioning token. On spontaneous non-English speech, the model falls back to its English training prior and produces output that reads like a real-time translation. This is a known architectural limitation.

This PR adds a post-processing step in the decoder that partially mitigates it for French.

## Mechanism

When the joint model picks a token from an English-exclusive blocklist (48 space-prefixed SentencePiece tokens: `' the'`, `' and'`, `' with'`, etc.), `applyEnglishBlocklist` scans the top-64 logits for the highest-scoring non-blocked Latin-script alternative and substitutes it. If nothing suitable is in top-64, the original prediction stands.

The substituted score is fed back into the transducer's LSTM state, which compounds: each forced French token nudges the following frames away from English.

Runs in O(64) per non-blank token. No additional model call, no latency impact.

## Benchmark

French recordings, drift = % sentences flagged by English function-word scan or langdetect=EN:

| Recording | Before | After |
|---|---|---|
| Children's weather segment | 0% | 0% |
| Archival 1912 | 2.9% | 2.7% |
| Weightlifting documentary | 7.1% | 0% |
| French slang documentary | 18.2% | 3.7% |
| Picard dialect documentary | 16.7% | 0% |
| Spontaneous interview (57 min) | 31.3% | 13.5% |

The remaining 13.5% on the interview is architectural: sustained English passages where no French candidate appeared in top-64 at all. The blocklist cannot help there.

## Test recordings

All INA recordings are publicly available on YouTube:

- [Argot et langage populaire (1981)](https://www.youtube.com/watch?v=zem3PVMsy8g)
- [Phonographe de 1912](https://www.youtube.com/watch?v=amPtXEaXQO0&t=12s)
- [Le parler picard (1982)](https://www.youtube.com/watch?v=W17uYuQA9Mc&t=8s)
- [Meteo pour les enfants (1980)](https://www.youtube.com/watch?v=sMSRyg172C0&t=8s)
- [Halterophilie (1975)](https://www.youtube.com/watch?v=E4xalDom8VY&t=2s)

The sixth recording (57-minute spontaneous interview) is private.

## Scope

This patch is French/English specific. The 48-token blocklist was derived empirically from the Parakeet TDT v3 SentencePiece vocabulary for French. The mechanism generalises to other language pairs but building equivalent blocklists requires the same vocabulary analysis and, critically, someone who can actually evaluate the output quality. I only speak French, so I only benchmarked French. I'm not going to claim results for languages I can't verify.

Flagging this as a starting point and leaving the generalisation to contributors who know the other languages and have audio to test against.

## Context

I'm not a professional developer. I found this problem while building a personal project on top of FluidAudio and did the diagnostic work: instrumented the decoder to log English token wins, measured that 84% of them had a French alternative in the top-64, ran before/after benchmarks across six recordings.

### DIFF
--- a/Sources/FluidAudio/ASR/Parakeet/SlidingWindow/TDT/Decoder/TdtDecoderV3.swift
+++ b/Sources/FluidAudio/ASR/Parakeet/SlidingWindow/TDT/Decoder/TdtDecoderV3.swift
@@ -37,6 +37,46 @@ internal struct TdtDecoderV3: Sendable {
     private let modelInference = TdtModelInference()
     // Parakeet‑TDT‑v3: duration head has 5 bins mapping directly to frame advances
 
+    // English-exclusive whole-word token IDs used by applyEnglishBlocklist.
+    // Space-prefixed SentencePiece tokens that are essentially impossible in French prose.
+    static let englishBlocklistIds: Set<Int> = [
+        506, 1502,  // ' the', ' The'
+        575, 1976,  // ' and', ' And'
+        2530,  // ' they'
+        1180, 3247,  // ' you', ' You'
+        1868,  // ' with'
+        1050, 7603,  // ' that', ' That'
+        1974, 5831,  // ' this', ' This'
+        1647,  // ' have'
+        3109,  // ' from'
+        924, 4020,  // ' was', ' Was'
+        4250,  // ' were'
+        1714,  // ' are'
+        4908,  // ' been'
+        4223,  // ' would'
+        6167,  // ' could'
+        2783,  // ' will'
+        4396,  // ' their'
+        3357,  // ' there'
+        4611,  // ' when'
+        3470,  // ' what'
+        6843,  // ' where'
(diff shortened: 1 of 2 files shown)


Matth-93:
Hi @Alex-Wengg and @BrandonWeng! Wanted to flag something about the failing CI check.

The **Japanese ASR Benchmark (CTC)** job is marked as failed, but looking at the actual benchmark output in the log, it passed cleanly:

```
CER: 9.07%
Status: Passed
```

The job failure appears to be a Node.js 20 deprecation warning in the GitHub Actions runner setup, unrelated to this patch. The English blocklist is applied to the TDT decoder to prevent it from drifting into English on non-English recordings. It has no effect on Japanese or any other language path.

Happy to rerun the job if that helps move the review forward. Let me know if you need anything else from me!

PR #634, opened 2026-05-24
Fix/word boost improvements

smdesai:
  
This PR is a series of measurement-driven improvements to the CTC vocabulary rescoring pipeline plus a build cleanup. Each change was validated against earnings22 (771 files, ~12K vocab terms) and a separat FDA / FDA-extended drug benchmarks.

The baseline earnings22 benchmarks at: TP 1133 / FP 6 / F 94.7  and with the PR, it's TP 1136 / FP 6 / F 94.8 It's a marginal improvement. 

The +1 frame correction is correct, not a tuning hack. It removes a known systematic bias in RNN-T decoders. The benchmark gain is small because earnings22 is fairly forgiving — the bias mattered most on a few edge cases. But every downstream consumer of TDT timestamps inherits the right numbers from now on: subtitle alignment, diarization-ASR alignment, streaming finalization, any future forced-alignment work. That's hard to capture in earnings22 numbers but real.

**TDT emission-delay correction and tighter rescoring margin**

Calibration-driven correctness fix.

Measured TDT-emitted token timestamps vs CTC log-prob argmax peaks across 8 667 word samples on full earnings22 (both v2 and v3). Median offset is consistently +1 encoder frame (~80 ms): TDT emits roughly one frame after the acoustic event, which biases the rescorer's CTC search window upstream of where the keyword actually lives.

Two changes:
1. Subtract one frame from emitted timestamps in createTokenTimings. Bounded with max(0, ...). Override via TDT_EMISSION_DELAY_FRAMES env var for sweep experiments.
2. Tighten defaultMarginSeconds from 0.5 → 0.10. Margin sweep showed identical TP/FP/F-score from 0.5 down to 0.10; first regression at 0.05 (-8 TP earnings22). 0.10 covers the post-correction p10..p90 of the offset distribution (~±5 frames).

**emission-delay-benchmark for measuring TDT vs CTC offsets**

New CLI subcommand that compares TDT-emitted timestamps against CTC argmax peaks on earnings22 (or any compatible dataset). Per-word constrained CTC DP within a configurable window; outputs aggregate stats (mean / median / p10..p90) plus per-sample JSON with TDT confidence and word duration for downstream analysis.

This is the calibration tool used to derive the +1-frame correction. The per-sample dump also drove the post-hoc analysis that confirmed adaptive correction has no useful headroom and that confidence isn't predictive of offset.

Flags:
  - --tdt-version (v2/v3/110m)
  - --search-half-width, --max-offset-seconds
  - --min-ctc-score, --min-word-chars
  - --max-files

  Combine with TDT_EMISSION_DELAY_FRAMES=0 to measure raw uncorrected offsets.
  
The emission-delay-benchmark tool is the real durable artifact. When future model updates change the encoder stride or the emission delay, somebody will need to remeasure. That tool means they don't have to rebuild it.

We measured an asymmetry that's documented in the RNN-T literature but hadn't been quantified for our specific models:
  TDT decoders tend to emit a token a few frames after the audio event that triggered it. The rescorer was building its
  CTC search window from TDT timestamps, so it was systematically looking upstream of where the acoustic evidence
  actually was.

The tool answered three questions:

1. Is the bias real? Yes. Median offsetStart = +1 frame (~80 ms) on full earnings22, consistent across v2 and v3. 8 667 word samples.
2. Is a static correction the right fix, or do we need a per-token model? Static is fine — bucketing the per-sample dump showed median is +1 across the bulk duration range; confidence is uncorrelated with offset.
3. What's the residual variance? ~3.5 frame stdev. That's noise we can't fix with a global shift, which informed the new tighter marginSeconds = 0.10 — wide enough to absorb the variance, narrow enough to avoid pulling in adjacent-word distractors.

Without the tool, the +1 frame correction would have been a guess. With it, the correction it's a measurement.


Sources/FluidAudioCLI/Commands/ASR/Parakeet/SlidingWindow/EmissionDelayBenchmark.swift
Alex-Wengg, line 1: whats the purpose of a emission delay benchmark ?

Sources/FluidAudio/ASR/Parakeet/SlidingWindow/CustomVocabulary/ContextBiasingConstants.swift
```diff
@@ -192,16 +196,62 @@ public enum ContextBiasingConstants {
...
         let isLarge = size > largeVocabThreshold
-        return VocabSizeConfig(
-            minSimilarity: isLarge ? 0.60 : 0.50,
-            cbw: isLarge ? 2.5 : 3.0
-        )
+        let minSimilarity: Float
+        if isExtraLarge {
+            minSimilarity = 0.60
+        } else if isLarge {
+            minSimilarity = 0.55
```
Alex-Wengg, line 250: whats with these hardcoding valuations

Sources/FluidAudio/ASR/Parakeet/SlidingWindow/CustomVocabulary/ContextBiasingConstants.swift
```diff
@@ -192,16 +196,62 @@ public enum ContextBiasingConstants {
         public let cbw: Float
     }
 
+    /// Threshold for classifying vocabulary as "extra-large".
```
Alex-Wengg, line 199: i think details like this should be in the  doc, coding agents should be able to get all the contextful or critical info from one source other wise they are more prone to bugs or confusions

Sources/FluidAudio/ASR/Parakeet/SlidingWindow/CustomVocabulary/Rescorer/VocabularyRescorer+TokenRescoring.swift
```diff
@@ -708,4 +783,197 @@ extension VocabularyRescorer {
...
+            // the spotter score, evaluated in its own (acoustic) window,
+            // beats the original stopword's CTC score in that same
+            // window — yet the spotter window is several frames wider
+            // than the lone TDT word it lands on.
+            let normalizedSpanWords =
+                span.map { Self.normalizeForSimilarity(wordTimings[$0].word) }
+            if span.count == 1 && Self.stopwords.contains(normalizedSpanWords[0]) {
+                continue
+            }
+            if span.count >= 2 {
```
Alex-Wengg, line 897: what does span count here do

Sources/FluidAudio/ASR/Parakeet/SlidingWindow/CustomVocabulary/WordSpotting/BpeTokenizer.swift
```diff
@@ -97,19 +97,35 @@ public final class BpeTokenizer: Sendable {
...
-        let preprocessed = "▁" + normalized.replacingOccurrences(of: " ", with: "▁")
+        let leading = prependWordBoundary ? "▁" : ""
+        let preprocessed = leading + normalized.replacingOccurrences(of: " ", with: "▁")
 
         // Split into characters
         var word = preprocessed.map { String($0) }
 
         // Apply BPE merges iteratively
-        while true {
+        while word.count >= 2 {
```
Alex-Wengg, line 129: why 2

Sources/FluidAudio/ASR/Parakeet/SlidingWindow/CustomVocabulary/WordSpotting/BpeTokenizer.swift
```diff
@@ -97,19 +97,35 @@ public final class BpeTokenizer: Sendable {
...
+        _ text: String,
+        addSpecialTokens: Bool = false,
+        prependWordBoundary: Bool = true
+    ) -> [Int] {
         // Normalize: lowercase + NFKC normalization (matches NeMo CTC models)
         let normalized = text.lowercased().precomposedStringWithCompatibilityMapping
 
         // Pre-tokenize: replace spaces with ▁ (sentencepiece style)
-        let preprocessed = "▁" + normalized.replacingOccurrences(of: " ", with: "▁")
+        let leading = prependWordBoundary ? "▁" : ""
```
Alex-Wengg: should we make "_" a constant, make it easier for context and system wide changes and debugging

Alex-Wengg (commented):
please make sure this script is also up to date 

https://github.com/FluidInference/FluidAudio/blob/main/Scripts/parakeet_subset_benchmark.sh#L8

Sources/FluidAudioCLI/Commands/ASR/Parakeet/SlidingWindow/EmissionDelayBenchmark.swift
smdesai: It's a calibration tool that measures how late TDT emits each token compared to where the corresponding acoustic event actually lives in the audio.

It solves the following: 
The rescorer builds its CTC search window from TDT timestamps. If TDT consistently emits tokens N frames after the acoustic event, the rescorer's window is centered N frames upstream of where the keyword's acoustic evidence actually is — degrading rescoring quality for reasons that are invisible without measurement.

This is a known property of RNN-T-style decoders (TDT included). NeMo even has a fastemit_lambda training-time knob to mitigate it. But the magnitude of the delay is model-specific, encoder-stride-specific, and quantization-specific. You can't just guess it; you have to measure it.

The tool answers three questions:
1. Is there a systematic bias? Yes/no, and how big.
2. Is a static correction sufficient, or do we need per-token logic? Bucket the offsets by confidence and word duration to see if the bias is uniform or varies.
3. What's the residual variance after the bias is removed? Tells you the right size for the rescorer's search window — wide enough to absorb noise, narrow enough to avoid distractors.

Sources/FluidAudio/ASR/Parakeet/SlidingWindow/CustomVocabulary/ContextBiasingConstants.swift
smdesai: Those are the string-similarity thresholds for the rescorer's TDT-anchored matching path, scaled by vocabulary size. They were derived from testing different datasets.

Vocabulary size changes the optimal threshold because of distractor density:
- Small vocab (≤ 10 terms, e.g. earnings22 with ~5-9 keywords per file): few terms, low chance any two are confusable. A permissive 0.50 threshold is fine — the small candidate set means a borderline match is unlikely to collide with another vocab term.
- Large vocab (≤ 100 terms, e.g. FDA with ~238 drug names): more terms, more chances for spurious near-matches. Raise to 0.55 to filter weaker matches.
- Extra-large vocab (> 100 terms, e.g. FDA-extended with ~670 terms including 624 distractor drugs that don't appear in the audio): high distractor density. Many drugs share roots (Vimza, Romvimza, Cimzia, etc.) so weak similarities catch the wrong term. Raise to 0.60.

These were chosen on dataset tests. On FDA-extended:
  - minSimilarity 0.55 → 33 false positives, F = 86.8
  - minSimilarity 0.60 → 8 false positives at -1 TP cost, F = 91.4

So the +0.05 raise eliminated 25 FPs at a cost of 1 TP — a clear precision/recall tradeoff favoring tightening when distractors are dense. Raising further didn't help (precision was already ceiling-bound).

Sources/FluidAudio/ASR/Parakeet/SlidingWindow/CustomVocabulary/ContextBiasingConstants.swift
smdesai: This is now documented in CustomVocabulary.md

Sources/FluidAudio/ASR/Parakeet/SlidingWindow/CustomVocabulary/Rescorer/VocabularyRescorer+TokenRescoring.swift
smdesai: span.count is the number of consecutive TDT-decoded words being considered as a single replacement candidate. It governs which stopword rule to apply.

The rescorer doesn't just match single words to vocabulary terms, it also tries multi-word spans. Two reasons:
1. Multi-word vocabulary terms. If the vocab contains "New York City" and TDT decoded "new york city", the rescorer matches a 3-word span against the term. Span count = 3.
2. TDT splits an unfamiliar word. If TDT decodes Romvimza as "rom vimza" (two words), the rescorer matches a 2-word span against the single vocab term Romvimza. Span count = 2.

Both branches reject replacement candidates whose original phrase is "all stopwords," but they differ in strictness for a reason:

Single-word case (span.count == 1): skip if the one word is a stopword. This prevents replacements like and → Andre, the → Theranos, is → Issa. Single-word stopword replacements are the highest-FP class because the spotter often finds some acoustic match for short common words against any reasonably similar vocabulary term.

Multi-word case (span.count >= 2): skip only if all words in the span are stopwords. The reasoning: if every word in at the is a stopword, replacing the whole thing with a vocab term is suspicious — the original was probably correct. But if even one word in the span is content (e.g. incident at the), the replacement might be legitimate (incident at the → Incidents-at-Heathrow is plausible if that's a vocab term).

The asymmetry — strict for single words, lenient for multi-word — exists because:
- Multi-word spans need acoustic evidence across multiple consecutive words to even be matched. The probability that a random multi-word stopword sequence fits a vocabulary term acoustically is much lower than for a single word, so the gate doesn't need to be as aggressive.
- Multi-word matches with content words mixed in are usually legitimate (e.g. new red → Newrez). Filtering them out by total stopword count would lose real recoveries.

Sources/FluidAudio/ASR/Parakeet/SlidingWindow/CustomVocabulary/WordSpotting/BpeTokenizer.swift
smdesai: The 2 is the structural minimum for BPE to do any work as a merge requires 2 adjacent symbols.

Sources/FluidAudio/ASR/Parakeet/SlidingWindow/CustomVocabulary/WordSpotting/BpeTokenizer.swift
smdesai: Yes it really ought to be as it occurs elsewhere in the codebase. I've updated the files accordingly.

Sources/FluidAudio/ASR/Parakeet/SlidingWindow/CustomVocabulary/Rescorer/VocabularyRescorer+TokenRescoring.swift
```diff
@@ -81,13 +106,31 @@ extension VocabularyRescorer {
...
+            if abs(simDiff) > 0.05 {
+                return simDiff > 0  // Prefer higher similarity
+            }
             if a.candidate.spanLength != b.candidate.spanLength {
-                return a.candidate.spanLength < b.candidate.spanLength  // Prefer shorter spans
+                return a.candidate.spanLength < b.candidate.spanLength  // Prefer shorter spans on near-tie
             }
-            // For same span length, prefer higher similarity
             return a.similarity > b.similarity
         }
```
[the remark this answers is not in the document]
smdesai: Code updated.

PR #653, opened 2026-06-01
feat(vad): FSMN-VAD backend (CoreML) [beta]

Alex-Wengg:
> [!NOTE]
> **Beta launch** — this backend ships as a beta: the public API surface and hosted model assets may still change in response to feedback before it is declared stable.

## Summary

Adds **FSMN-VAD** (FunASR, ~5.2M) as a CoreML voice-activity-detection backend. Model: [`FluidInference/fsmn-vad-coreml`](https://huggingface.co/FluidInference/fsmn-vad-coreml).

## Pipeline

```
waveform → [Preprocessor fp32/CPU] → 400-d features (fbank80 + LFR m=5,n=1)
        → [FSMN fp16/ANE, enumerated 512..3072] → [1,T,248] frame scores (col 0 = silence prob)
        → host decision → speech segments [start_ms, end_ms]
```

The **decision** ports FunASR's `FsmnVADStreaming`: per-frame `speech if silence_prob ≤ 0.2`, a 20-frame sliding-window hysteresis (sil→speech / speech→sil at 15), silence→endpoint after `max_end_silence` (800 ms), `lookback`/`lookahead`, and `max_single_segment` (60 s). Audio > ~30 s is processed in chunks; per-frame silence probs are concatenated and the decision runs once.

(RangeDim breaks the FSMN's dilated conv on the BNNS path, so the scorer uses fixed enumerated buckets and the host chunks long audio.)

## Changes

- `ModelNames`: `fsmnVad` `Repo` + `FsmnVad` registry
- `Sources/FluidAudio/VAD/Fsmn/`: `FsmnVadModels`, `FsmnVadManager` (+ `FsmnVadSegment`)
- CLI: `fsmn-vad-segment`

## Verification

On a 20 s clip: `[120, 19960]` vs FunASR `[70, 19980]` — boundaries within ~50 ms. 0.04 s, 0.056 GB peak.

## Notes

- Alternative to FluidAudio's existing `silero-vad`.
- The decision is faithful-to-behavior (matches FunASR within ~50 ms), not a bit-exact port of the full streaming state machine.


LemonCANDY42:
Thanks for adding the FSMN-VAD backend. I’m curious whether there are any existing benchmark comparisons between Ten-VAD and FSMN-VAD for this use case, especially around VAD accuracy and streaming performance/latency. If you’ve already looked at Ten-VAD, I’d be interested to know how it compares as a CoreML VAD candidate.

- https://github.com/ten-framework/ten-vad

Alex-Wengg:
@LemonCANDY42 i have not had the time to benchmark. i am too sure how it performs compared to tad but i think vad silero is still the bset

LemonCANDY42:
> @LemonCANDY42 i have not had the time to benchmark. i am too sure how it performs compared to tad but i think vad silero is still the bset

Got it, thanks.

Sources/FluidAudio/VAD/Fsmn/FsmnVadManager.swift
```diff
@@ -0,0 +1,163 @@
...
+        let chunkSamples = (Self.buckets.last! - Self.windowFrames) * 160
+        var sil: [Float] = []
+        var offset = 0
+        while offset < audio.count {
+            let end = min(offset + chunkSamples, audio.count)
+            let chunk = Array(audio[offset..<end])
+            // Drain CoreML's autoreleased MLMultiArrays per chunk so memory stays
+            // bounded on long audio; otherwise they accumulate across the whole file.
+            let chunkSil = try autoreleasepool { try chunkSilence(chunk) }
+            sil.append(contentsOf: chunkSil)
```
LauraGPT, line 74: The long-audio chunking currently compresses the output timeline. From the published FsmnVadPreprocessor.mlmodelc/model.mil, each chunk uses a valid 400-sample frame convolution with stride 160, prepends two fbank frames, then applies a valid width-5 LFR convolution. Therefore the current 488,320-sample chunk produces:

- floor((488320 - 400) / 160) + 1 = 3050 fbank frames
- 3050 + 2 - 5 + 1 = 3048 output frames

The loop consumes 30,520 ms but appends only 30,480 ms to sil; decide then maps the concatenated index back with frame * 10, losing 40 ms at every boundary. A 60-minute input has 118 chunks and ends up about 4.72 seconds early. It can also miss or perturb speech crossing a boundary because the discarded frontend context is not overlapped.

Please preserve absolute chunk offsets or overlap the frontend context and trim duplicate outputs before concatenation. A regression with speech or impulses immediately before and after at least two chunk boundaries should assert absolute segment timestamps against the FunASR reference, and ensure drift does not grow with chunk count. The existing 20 s fidelity check never enters this path.

LauraGPT (commented):
Found one correctness issue in the long-audio path; details inline.

PR #688, opened 2026-06-11
fix(asr): splice long-form chunk merges on SentencePiece word boundaries

### LINKED ISSUE #683: Long-form chunk merge can glue words across window seams: LCS/midpoint merge never checks SentencePiece word boundaries (`▁` prefix)
### Summary

On long-form transcription (audio > ~15 s, the stateless `ChunkProcessor` path), the merged transcript intermittently glues the last word of one window to the first word of the next ("wordone wordtwo" → "wordonewordtwo"). We see this consistently with `parakeet-tdt-0.6b-v3-coreml` on Ukrainian dictation; frequency increases with utterance length. Present in 0.14.1 and still reproducible in 0.15.2 (the rewritten merge reduced but did not eliminate it).

### Mechanism

Tokens are SentencePiece pieces where a leading `▁` marks a word start. At a window seam the same audio is tokenized independently by both windows, and the two decoders frequently disagree on segmentation near the boundary (the right window often emits its first word WITHOUT the `▁` prefix, since for that decoder the word is utterance-initial or split differently).

`ChunkProcessor.mergeChunks` (Sources/FluidAudio/ASR/Parakeet/SlidingWindow/TDT/ChunkProcessor.swift, ~675) stitches `left` and `right` token streams via:

1. timestamp-matched pairs → `mergeUsingMatches` (~778), or
2. LCS fallback (`SequenceMatcher.findLongestCommonSubsequence`, ~746) → `mergeUsingMatches`, or
3. `mergeByMidpoint` (~822) when no matches exist.

None of the three paths validates that the splice point lands on a word boundary: when the merge switches from `left[...]` to `right[(lastRight + 1)...]` (or fills gaps with `gapRight`), the first token taken from `right` can be a continuation piece (no `▁`). Decoded text then concatenates it directly onto the last `left` word — a glued word with no space. `mergeByMidpoint`'s pure time-cutoff filter has the same blind spot.

### Reproduction

1. `parakeet-tdt-0.6b-v3-coreml`, default `ASRConfig` (also reproduces with `melChunkContext: false`).
2. Transcribe continuous speech > 15 s (so at least one window seam exists). Non-English (we use Ukrainian) makes it more frequent, but the mechanism is language-independent.
3. Inspect decoded text around the ~15 s / ~30 s marks: intermittently two words are concatenated without a space; occasionally a hybrid word appears (left's word tail + right's continuation pieces).

A targeted unit repro: feed `mergeUsingMatches` a `right` stream whose post-match tail starts with a non-`▁` token while `left`'s last kept token ends a different word — the merged sequence decodes to a glued word. (`mergeTokenWindowsForTesting` at ~376 makes this easy to drive.)

### Suggested direction

When splicing tokens from `right` (post-match tail in `mergeUsingMatches`, gap fills, and the `mergeByMidpoint` cutoff), verify the first appended `right` token is word-initial (`▁` prefix or punctuation); if not, shift the splice point left/right to the nearest word boundary, or prefer the side that keeps the word intact. This keeps the merge language-agnostic — it inspects only the tokenizer's own word-boundary marker, not text.

### Environment

- FluidAudio 0.15.2 (7f963cdc), also observed on 0.14.1
- macOS 15 / iOS 18+26, `parakeet-tdt-0.6b-v3-coreml`
- Integration: AsrManager.transcribe(samples, decoderState:, language:) with >15 s buffers

Alex-Wengg:
Fixes #683.

## Problem

`ChunkProcessor`'s window merge splices token streams purely by token IDs and timestamps. None of the three merge paths checks SentencePiece word boundaries, so a seam splice could decode:

- a **hybrid word** — left window's word prefix + right window's continuation pieces,
- a **glued word** — a continuation piece (no `▁` prefix) appended directly onto a completed word, or
- a **truncated word** — a raw time cutoff splitting a word.

## Fix

- `ChunkProcessor.spliceSafeTokenIds(vocabulary:)` derives the set of word-initial token IDs (`▁`/space-prefixed or punctuation-only pieces) from the model vocabulary, computed once per merge pass and threaded through `mergeChunks`.
- **`mergeUsingMatches` post-match tail**: when the splice lands mid-word (first tail token is a continuation piece), adopt the **right window's segmentation of the whole seam word** — the left chunk is typically the one whose audio was cut mid-word. Only when the right stream itself starts mid-word (no `▁` piece before the matched anchor — the construction that produces glued words) does the merge keep the left window's word and resume right at its next word-initial piece.
- **`mergeByMidpoint`**: extends the left stream to finish the word the time cutoff splits, and drops orphaned continuation pieces from the right stream's head.
- Without a vocabulary the merge is byte-for-byte unchanged.

Note on the issue's analysis: gap fills turned out to be structurally safe — splices between matches are always anchored at identical token IDs, so they stay stream-contiguous. Only the post-match tail and the midpoint cutoff can splice mid-word.

The issue's suggested repair (complete the seam word from the *left* window) was implemented first and rejected empirically: on FLEURS uk_ua it truncated 4 seam words (e.g. `Гренландією`→`Гренланді`, `воскресіння`→`воскресен`) because at a seam the left window is the one cut mid-word — the right window heard the word in full.

## Validation

- **1h earnings22 audio (~277 window seams)**: the only 3 word-level diffs vs baseline are organic seam-splice artifacts repaired by this change — `Patrict.`→`Patrick.` (seam 47), `outlocate`→`allocate` (seam 144), `automoti`→`automotive` (seam 274). Zero regressions.
- **FLEURS uk_ua, 200 samples**: WER 7.0% == baseline (CER 2.3%); the 28 files ≥ 15s (every one with a seam) are WER-neutral with no truncated or hybrid seam words.
- **FLEURS en_us, 200 samples**: WER 5.0% == baseline. RTFx unchanged (~140x).
- New unit tests cover the glue repro from the issue, the adopt-right-segmentation path, the right-cut-mid-word fallback, the midpoint path, piece classification, and unchanged legacy behavior without a vocabulary.

### DIFF
--- a/Documentation/ASR/LongTranscription.md
+++ b/Documentation/ASR/LongTranscription.md
@@ -213,7 +213,7 @@ Notes for tuning:
 After each chunk decodes independently, `ChunkProcessor.mergeChunks` stitches
 adjacent chunks into a single token timeline. The merger never re-runs the
 decoder and never invents tokens; it only chooses which side of the overlap
-each token comes from. The strategy is a three-step ladder:
+each token comes from. The strategy is a four-step ladder:
 
 1. **Disjoint shortcut.** If the left chunk's last token ends before the
    right chunk's first token begins, concatenate without merging.
@@ -234,6 +234,35 @@ cannot collapse two different words that happen to share a substring, and it
 is robust to small per-chunk timestamp jitter. The contiguous-match path
 preserves order strictly; LCS is only entered when adjacent chunks disagree
 enough that a contiguous run would be dishonest.
+
+### Word-Boundary Splice Repair
+
+Matching alone does not make the *splice points* safe: the two windows
+tokenize the same seam audio independently, and the right window often
+segments its first word differently (frequently without the SentencePiece
+`▁` word-start prefix, since for that decoder the word is utterance-initial).
+Splicing a continuation piece from the right stream directly after a left
+token decodes as a glued or hybrid word — `"work" + "ks"` → `"worksks"`
+(issue #683).
+
+To prevent this, `mergeChunks` derives a set of *splice-safe* token IDs from
+the model vocabulary once per merge pass: pieces with a `▁`/space prefix or
+punctuation-only pieces. Splices are then repaired at word granularity:
+
+- **`mergeUsingMatches`** — when the post-match tail of the right stream
+  starts with a continuation piece, the merger adopts the right window's
+  segmentation of the entire seam word (the left chunk is typically the one
+  cut mid-word). Only when the right stream itself begins mid-word — the
+  construction that produced glued words — does the left window keep the
+  seam word, with the right stream resuming at its next word-initial piece.
+- **`mergeByMidpoint`** — the time cutoff is adjusted so the left stream
+  finishes the word it started, and orphaned continuation pieces (whose
+  word-initial piece was trimmed away) are dropped from the right stream's
+  head.
+
+The repair inspects only the tokenizer's own word-boundary marker, never
+transcript text, so it is language-agnostic. Without a vocabulary the merge
+is byte-for-byte unchanged.
 
 ## Streaming Threshold for Large Files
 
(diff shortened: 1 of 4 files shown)


holovchenko:
Hello @Alex-Wengg. I've checked it in the Fork, and it dramatically better on Ukrainian. Thank you!

Issue #691, opened 2026-06-12
KokoroAne English frontend over-stresses function word "to" as tˈO


LemonCANDY42:
## Summary

`KokoroAneManager`'s English text frontend currently phonemizes text by whitespace-splitting into words, running `G2PModel.shared.phonemize(word:)` per word, and joining the IPA parts. This maps the common function word `to` to `tˈO`. With the default English voice (`af_heart`), this is rendered as a stressed long "toe" / tour-like pronunciation in ordinary sentences such as "I want to go".

This looks like an English frontend issue rather than a CoreML acoustic-chain issue: bypassing the text frontend and feeding a weak form such as `tə` through `synthesizeFromPhonemes(...)` avoids the bad pronunciation.

## Environment

- FluidAudio: `0.14.7` (`8048812869b0c7c6fa393e564a4fb6f95126ba23`)
- Backend: `KokoroAneManager(variant: .english, defaultVoice: "af_heart")`
- Model assets: `FluidInference/kokoro-82m-coreml`, English `ANE/` 7-stage chain plus shared `kokoro/G2PEncoder.mlmodelc`, `G2PDecoder.mlmodelc`, `g2p_vocab.json`
- Platform reproduced on macOS host tests for G2P output; audible symptom observed in iOS app using the same `af_heart` default voice path

## Minimal G2P repro

A small test using the current internal `G2PModel` gives:

```text
word=to   ipa=tˈO
word=too  ipa=tˈu
word=two  ipa=tˈu
word=tour ipa=tˈʊɹ
```

The issue is not that `to` literally becomes `tour` (`tour` includes `ʊɹ`), but that `to` receives a stressed diphthong form (`tˈO`) in contexts where English normally expects a weak function-word pronunciation (`tə`, sometimes `tu`).

## Relevant code path

`Sources/FluidAudio/TTS/KokoroAne/KokoroAneManager.swift`:

```swift
let words = trimmed.split(whereSeparator: { $0.isWhitespace }).map(String.init)
...
let cleaned = word.trimmingCharacters(in: .punctuationCharacters).lowercased()
if let ipa = try await G2PModel.shared.phonemize(word: cleaned) {
    parts.append(ipa.joined())
}
...
return parts.joined(separator: " ")
```

Because this is per-word and context-free, `to` gets the same stressed form everywhere.

## Upstream comparison

[...]

PR #696, opened 2026-06-13
fix(tts): treat KokoroAne quote delimiters as punctuation

### LINKED ISSUE #695: KokoroAne English phonemizer treats quoted words as OOV
Follow-up to #691 and #692.

## Problem

The merged Misaki-lexicon-first frontend fixes ordinary words such as `to`, but ASCII single quotes can still be kept inside the word token. For example, an input like `'to'` is tokenized as a word-like token instead of `to` plus quote delimiters. The lexicon then misses and the word can fall back to BART G2P, reintroducing the over-pronounced/OOV behavior that #692 was meant to avoid.

This matters for chat/TTS text because quoted example words are common, e.g. `Let's practice 'to'`.

## Expected

Outer quote delimiters should not be part of the lexicon key. `'to'` should resolve through the same Misaki lexicon entry as `to`, while contractions such as `there's` and `don't` should remain intact.

## Proposed fix

In `KokoroAneEnglishPhonemizer.splitWords`, treat ASCII apostrophe as a word character only when it is internal to a word: there is already current word content and the following character is a letter or digit. Otherwise emit it as punctuation so the existing unsupported-punctuation path drops it.

I have a small follow-up PR prepared with regression coverage for `'to'` and a quoted sentence containing `there's`.

LemonCANDY42:
Fixes #695.
Follow-up to #691 and #692.

## Summary

- Treat ASCII apostrophe as a KokoroAne English word character only when it is internal to a word.
- Preserve contractions such as `there's` and `don't`.
- Let quoted words such as `'to'` resolve through the same Misaki lexicon key as `to` instead of falling through to BART G2P.
- Add regression tests for quoted `to` and a quoted sentence containing `there's`.

## Root cause

After #692, lexicon-first lookup fixes ordinary `to`, but `splitWords` still kept every ASCII apostrophe inside the current word token. A quoted example word like `'to'` could therefore miss the Misaki lexicon and fall back to BART G2P.

## Validation

- `swift build --target FluidAudio`
- `git diff --check -- Sources/FluidAudio/TTS/KokoroAne/G2P/English/KokoroAneEnglishPhonemizer.swift Tests/FluidAudioTests/TTS/KokoroAne/KokoroAneEnglishPhonemizerTests.swift`

`swift test --filter KokoroAneEnglishPhonemizerTests` currently does not complete on `main` in this checkout because the package build hits an existing CLI compile error in `Sources/FluidAudioCLI/Commands/ASR/Parakeet/Streaming/NemotronMultilingualFleursBenchmark.swift` (`the compiler is unable to type-check this expression in reasonable time`).

### DIFF
--- a/Sources/FluidAudio/TTS/KokoroAne/G2P/English/KokoroAneEnglishPhonemizer.swift
+++ b/Sources/FluidAudio/TTS/KokoroAne/G2P/English/KokoroAneEnglishPhonemizer.swift
@@ -145,6 +145,10 @@ struct KokoroAneEnglishPhonemizer: Sendable {
 
     // MARK: - Word splitter
 
+    private static let knownLeadingApostropheWords: Set<String> = [
+        "'cause", "'em", "'til", "'tis", "'twas", "'twere",
+    ]
+
     /// Emit runs of letters/digits (internal apostrophes and hyphens stay
     /// inside words: `don't`, `twenty-one`), single punctuation chars as
     /// their own tokens, and drop whitespace. Same shape as the StyleTTS2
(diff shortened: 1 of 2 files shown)


Alex-Wengg:
please fix the swift format issue

PR #749, opened 2026-07-02
fix: preserve model cache when first load is cancelled

ComicBit:
## Problem

`DownloadUtils.loadModels` treats **any** first-load failure as cache corruption: it deletes the whole repo cache folder and re-downloads. But cancellation is not corruption — if the load is cancelled (app teardown, user abort, `Task` cancellation propagating as `CancellationError` or `NSURLErrorCancelled` -999), the fallback wipes a perfectly valid, fully-downloaded model cache. We hit this in production: cancelling a Sortformer load during app stop deleted a 246 MB downloaded model, forcing a full re-download on next launch.

## Fix

Add an `isCancellationError` classifier and check it before the delete-and-redownload fallback. Cancelled loads rethrow without touching the cache. Detected shapes:

- Swift `CancellationError`
- `NSURLErrorDomain` code `NSURLErrorCancelled` (-999)
- `NSCocoaErrorDomain` code `NSUserCancelledError`
- any of the above anywhere in the `NSUnderlyingErrorKey` chain (depth-capped)

Offline-mode behaviour (`enforceOffline`) is unchanged — it already skipped the wipe.

## Tests

`Tests/FluidAudioTests/Shared/DownloadUtilsCancellationTests.swift` — 10 cases covering all detected cancellation shapes plus negatives (timeout, corrupt-file error, -999 in a non-URL domain, chain termination).

`swift build` + `swift test --filter DownloadUtilsCancellationTests` green on macOS.

### DIFF
--- a/Sources/FluidAudio/DownloadUtils.swift
+++ b/Sources/FluidAudio/DownloadUtils.swift
@@ -269,8 +269,9 @@ public class DownloadUtils {
 
     /// `true` when `error` represents cancellation rather than a corrupted
     /// cache: Swift `CancellationError`, `NSURLErrorCancelled` (-999), or
-    /// `NSUserCancelledError` — including anywhere in the underlying-error
-    /// chain (`NSUnderlyingErrorKey`).
+    /// `NSUserCancelledError` — checked on the error itself and on the
+    /// `NSUnderlyingErrorKey` chain. The walk is depth-capped at 8 links as
+    /// cycle protection (real-world wrapping is 1–2 levels deep).
     ///
     /// Used by ``loadModels(_:modelNames:directory:computeUnits:variant:progressHandler:)``
     /// to skip the delete-cache-and-redownload fallback for cancelled loads.
--- a/Sources/FluidAudio/DownloadUtils.swift
+++ b/Sources/FluidAudio/DownloadUtils.swift
@@ -231,6 +231,15 @@ public class DownloadUtils {
                 throw error
             }
 
+            // Cancellation is not corruption. A cancelled first load (app
+            // teardown, user abort) must never wipe a valid cache — deleting
+            // here threw away fully-downloaded multi-hundred-MB repos.
+            if isCancellationError(error) {
+                logger.info(
+                    "Load cancelled; preserving model cache. \(error.localizedDescription)")
+                throw error
+            }
+
             logger.warning("First load failed: \(error.localizedDescription)")
             logger.info("Deleting cache and re-downloading…")
             let repoPath = directory.appendingPathComponent(repo.folderName)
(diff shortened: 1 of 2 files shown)


Sources/FluidAudio/DownloadUtils.swift
```diff
@@ -258,6 +267,31 @@ public class DownloadUtils {
         }
     }
 
+    /// `true` when `error` represents cancellation rather than a corrupted
+    /// cache: Swift `CancellationError`, `NSURLErrorCancelled` (-999), or
+    /// `NSUserCancelledError` — including anywhere in the underlying-error
+    /// chain (`NSUnderlyingErrorKey`).
```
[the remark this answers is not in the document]
ComicBit: Done in 7c4d2063 — doc now states the 8-link depth cap and the rationale (cycle protection; real-world wrapping is 1–2 levels deep).

Tests/FluidAudioTests/Shared/DownloadUtilsCancellationTests.swift
```diff
@@ -0,0 +1,80 @@
...
+
+    func testUnderlyingChainCycleTerminates() {
+        // Self-referential underlying chain must not loop forever; the
+        // walker is depth-capped.
+        let inner = NSError(domain: "com.example.a", code: 1)
+        let outer = NSError(
+            domain: "com.example.b", code: 2,
+            userInfo: [NSUnderlyingErrorKey: inner])
+        XCTAssertFalse(DownloadUtils.isCancellationError(outer))
+    }
```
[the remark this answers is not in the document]
ComicBit: Done in 7c4d2063 — renamed to `testNonCancellationUnderlyingChainNotClassified` (it never built a cycle), and added `testDepthCapTerminatesLongChain` which actually exercises the cap on a 12-link chain.

Sources/FluidAudio/DownloadUtils.swift
```diff
@@ -258,6 +267,32 @@ public class DownloadUtils {
         }
     }
 
+    /// `true` when `error` represents cancellation rather than a corrupted
```
Alex-Wengg: the comment is abit too verbose , if users or agents want more context they will use commit history as context reference.

Sources/FluidAudio/DownloadUtils.swift
```diff
@@ -258,6 +267,32 @@ public class DownloadUtils {
...
+    /// cycle protection (real-world wrapping is 1–2 levels deep).
+    ///
+    /// Used by ``loadModels(_:modelNames:directory:computeUnits:variant:progressHandler:)``
+    /// to skip the delete-cache-and-redownload fallback for cancelled loads.
+    static func isCancellationError(_ error: Error) -> Bool {
+        if error is CancellationError { return true }
+
+        var current: NSError? = error as NSError
+        var depth = 0
+        while let nsError = current, depth < 8 {
```
Alex-Wengg: why specifically 8 , also ideally this should be a constant somewhere

Sources/FluidAudio/DownloadUtils.swift
ComicBit: Trimmed in 243a591a — doc comment is now three lines; the depth-cap rationale moved onto the constant itself.

Sources/FluidAudio/DownloadUtils.swift
ComicBit: Done in 243a591a — extracted `maxUnderlyingErrorDepth = 8` with a one-line rationale (cycle guard; real-world wrapping is 1–2 levels deep; 8 leaves generous headroom).

Sources/FluidAudio/DownloadUtils.swift
```diff
@@ -267,20 +267,19 @@ public class DownloadUtils {
...
-    /// `true` when `error` represents cancellation rather than a corrupted
-    /// cache: Swift `CancellationError`, `NSURLErrorCancelled` (-999), or
-    /// `NSUserCancelledError` — checked on the error itself and on the
-    /// `NSUnderlyingErrorKey` chain. The walk is depth-capped at 8 links as
-    /// cycle protection (real-world wrapping is 1–2 levels deep).
-    ///
-    /// Used by ``loadModels(_:modelNames:directory:computeUnits:variant:progressHandler:)``
-    /// to skip the delete-cache-and-redownload fallback for cancelled loads.
+    /// Guards the `NSUnderlyingErrorKey` walk against pathological or cyclic
+    /// error chains; real-world wrapping is 1–2 levels deep.
```
Alex-Wengg: so why do we need 8 then ? why not just 2

ComicBit: The depth limit was pretty arbitrary. In reality, NSError chains are usually only 1–2 levels deep and a plain NSError can't even form a cycle because userInfo is immutable

PR #751, opened 2026-07-03
fix(offline-diarizer): re-embed zero-vote spans instead of arbitrary cluster-0 tie-break

ComicBit:
## Problem

In `OfflineReconstruction.buildSegments`, aggregated timeline frames whose per-cluster vote sums are **all zero** are tie-broken arbitrarily to cluster 0. A frame ends up with zero votes when the active local speaker slot received no embedding in any covering window (assignment −2 everywhere). The result: whole speaker turns silently absorbed into the surrounding speaker's segment.

Reproduced on a 97 s two-speaker fixture: a clean 1.6 s turn (segmentation gaps delimit it correctly at 26.88–28.46 s, verified from raw powerset activations) had zero cluster votes in all six covering windows and was emitted as part of the other speaker's 13 s segment.

## Fix

Optional post-pass (`OfflineDiarizerConfig.zeroVoteReembed`, **disabled by default** — no behavior change unless opted in):
- detect maximal contiguous zero-vote runs (speech-active frames, no cluster votes, bounded by gaps/voted frames) ≥ `minDurationSeconds` (default 0.4 s)
- re-embed each run's exact audio span (zero-padded window, weight mask covering only the span's frames so neighboring audio can't leak in)
- assign to the closest speaker centroid by cosine; a failed/NaN embedding falls back to the existing tie-break
- the run becomes its own segment when its assigned cluster differs from its neighbors

## Results (fixture A/B, community-1 preset, min-segment 0.5)

| config | DER | speaker-error |
|---|---|---|
| baseline | 0.0404 | 0.0200 |
| + zeroVoteReembed | **0.0216** | **0.0012** |

Three real turns recovered (26.9–28.4 s, 72.6–73.7 s, 87.3–87.9 s), all matching ground truth.

## Tests

`ZeroVoteReembedderTests`: 15 cases — run detection (voted/non-speech bounds, min-duration filter, multiple runs, timeline-end), assignment (no-margin win, tie → lowest index, NaN, dimension mismatch), disabled-by-default, and synthetic-frame reconstruction integration (segment split on run boundaries, disabled path never invokes the embedder, nil-embedding parity with disabled).

Pure decision logic lives in `ZeroVoteReembedder` (no RNG, stable iteration order) so it is testable without CoreML models.

### DIFF
--- a/Sources/FluidAudio/Diarizer/Offline/Core/OfflineDiarizerManager.swift
+++ b/Sources/FluidAudio/Diarizer/Offline/Core/OfflineDiarizerManager.swift
@@ -321,11 +321,34 @@ public final class OfflineDiarizerManager {
             logger.debug("Clustering completed in \(clusteringTime)s with no assignments")
         }
 
+        // Zero-vote runs carry no clustering evidence, so reconstruction re-embeds their
+        // exact audio span to pick a speaker. Extraction needs models + audio, which the
+        // reconstruction helper doesn't own — hand it a closure instead.
+        let spanEmbedder: ((Double, Double) -> [Float]?)?
+        if config.zeroVoteReembed.enabled, !centroids.isEmpty {
+            let extractor = OfflineEmbeddingExtractor(
+                fbankModel: models.fbankModel,
+                embeddingModel: models.embeddingModel,
+                pldaTransform: pldaTransform,
+                config: config
+            )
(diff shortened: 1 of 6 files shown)


Alex-Wengg:
@ComicBit please fix the swift format and we can merge

ComicBit:
@Alex-Wengg sshould be good to merge. 🙏

PR #753, opened 2026-07-04
feat(diarizer): optional progressHandler on performCompleteDiarization

thatbin:
## What

Adds an optional `progressHandler: ((Double) -> Void)? = nil` to `DiarizerManager.performCompleteDiarization`. It reports overall progress in `0.0...1.0` after each processed audio chunk, ending at `1.0`.

## Why

`performCompleteDiarization` already loops over the audio in chunks, but exposes no way to observe progress, so a host app can only show an indeterminate spinner (or a fake full bar) during a long offline diarization. This lets callers drive a real progress bar.

## Notes

- Default `nil` — every existing call site is unaffected (fully source-compatible).
- The closure is non-escaping and invoked synchronously on the calling thread, so it adds no `Sendable`/concurrency requirements.
- The progress fraction is derived from the existing `stepSize` stride over `totalSamples`; no behavioral change to segmentation, embedding, or clustering.
- `swift format lint` clean; `swift build` green.

### DIFF
--- a/Sources/FluidAudio/Diarizer/Core/DiarizerManager.swift
+++ b/Sources/FluidAudio/Diarizer/Core/DiarizerManager.swift
@@ -130,6 +130,9 @@ public final class DiarizerManager {
     ///   - samples: Audio samples (16kHz mono) - accepts any RandomAccessCollection of Float
     ///             (Array, ArraySlice, ContiguousArray, or custom collections)
     ///   - sampleRate: Sample rate (default: 16000)
+    ///   - progressHandler: Optional callback invoked after each processed audio chunk with
+    ///             overall progress in the range `0.0...1.0`. Called synchronously on the
+    ///             calling thread; the final call reports `1.0` once processing completes.
     /// - Returns: `DiarizationResult` containing:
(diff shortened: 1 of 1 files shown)


Alex-Wengg:
@thatbin have you tested this in an app of yours

thatbin:
Yes, tested in an app. Pointed the SPM dependency at this branch and passed a `progressHandler` into `performCompleteDiarization` to drive a real progress bar. Diarization runs inside an actor for us, and since the closure is invoked synchronously on the calling thread, forwarding the fraction straight out was trivial — no extra `Sendable`/concurrency work:

```swift
let result = try manager.performCompleteDiarization(samples) { fraction in
    // fires after each processed chunk, 0.0...1.0, ends at 1.0
    progress(fraction)   // forwarded out of the diarization actor to the UI
}
```

Default `nil` kept every existing call site untouched. On a long multi-speaker recording the bar advances per chunk and settles at `1.0`, instead of the hardcoded full bar we had before — short capture:


https://github.com/user-attachments/assets/0d61d535-4206-4234-bade-90ad9e018528

Issue #758, opened 2026-07-05
Long-form transcription deterministically drops ~7 s of clear speech at a chunk boundary (95 s repro WAV linked)


maboa:
## Summary

`AsrManager.transcribe` on the long-form path (ChunkProcessor, 15 s chunks / 13 s stride) deterministically loses multi-second spans of clearly audible speech at chunk boundaries. On the linked 95-second clip, ~7 s of speech is dropped entirely — reproducibly, byte-identical output across runs.

## Environment

- FluidAudio 0.15.4 (exact-version pin), `parakeet-tdt-0.6b-v3` via `AsrModels.downloadAndLoad(version: .v3)`
- `AsrManager` with `ASRConfig.default`, called from a macOS app (Apple Silicon, macOS 26)
- Input: 16 kHz mono PCM WAV, so no resampling/decode variability

## Repro

1. Download and transcribe [`chunk-seam-drop-1235s.wav`](https://lab.hyperaud.io/temp/chunk-seam-drop-1235s.wav) (95 s, 16 kHz mono, conference Q&A audio, English).
2. Observe the output around t = 71.6–78.8 s.

**Expected** (what's audibly in the clip, and what the model itself produces for this span when the audio is windowed differently):

> …expectations from different communities? **Yes, I I can s uh give you some examples that will be uh uh uh** more in depth with the next talk too…

**Actual:** the last word before the gap is "communities?" (ends ~71.6 s); the next emitted word is "more" at 78.8 s. Everything between — the entire start of the answer — is missing. Note the gap terminates at 78.0 s = 6 × 13 s, i.e. exactly on the chunk-stride grid.

## What we've established (fuller data in https://github.com/hyperaudio/glider/issues/83)

The clip is cut from a ~30-min recording that exhibits 2–3 such drops plus occasional boundary garbles ("language in" → "languag ines"). Things we verified while investigating:

1. **Fully deterministic**: repeated runs produce word- and timestamp-identical output, so these are reproducible merge decisions, not sampling noise.
2. **`melChunkContext` relocates the failures rather than fixing them**: with `melChunkContext: false` the full file's three drops are all fixed — and a different ~7 s span drops at a different boundary instead, plus a new artifact class of duplicate token pairs emitted ~80 ms apart ("Okay Okay", "check check"). Neither mode dominates.
3. **`dualDecodeArbitration: true` was a no-op** on this file beyond its processing overhead — output byte-identical to plain `melChunkContext: false`.
[...]

Issue #706, opened 2026-06-16
Chunk merging in Parakeet Unified leaves artifacts


nub235:
For the offline version, it splits the audio into 15-second chunks with 2-second overlap and then tries to merge them together seamlessly, but when I'm using it, I can sometimes tell when it's doing the chunk merging because a word is duplicated. And then, like, the second instance is capitalized, or like there's just a word that's capitalized in the middle of a sentence. Would it be possible to improve the chunk merging logic so it doesn't leave visible artifacts?

Alex-Wengg:
Do you have testing examples

PR #800, opened 2026-07-16
fix(asr): end-align the final window at the last speech-bearing frame (#747)

### LINKED ISSUE #747: Trailing words silently dropped on quiet long-form audio (mel-context final-window blanks out)
## Summary

On the default `parakeet-tdt-0.6b-v3` batch path (`melChunkContext = true`), the **final** sliding window of a long (multi-window) utterance can decode to **all-blank** even though its audio contains clear speech — so the last several words are silently dropped from the transcript. It's intermittent and correlates with **low input gain** and a **pause** shortly before the dropped span. The same audio transcribes correctly when that tail is fed as a single window, or with `melChunkContext = false`, or when the input is amplified — which points at the 80 ms mel-context prepend (PR #264) degrading a short, low-amplitude final window's encoder features below the joint's blank threshold.

It's a *silent* failure (no error, high overall confidence), so it's easy to miss.

## Environment

- FluidAudio 0.15.4 line, `AsrModelVersion.v3` (`parakeet-tdt-0.6b-v3-coreml`), Apple Neural Engine, macOS 14.
- Default `ASRConfig` (`melChunkContext = true`, `dualDecodeArbitration = false`), batch `AsrManager.transcribe`.
- Observed via a fork that only adds an *orthogonal* encoder-feature accessor; the `ChunkProcessor` / mel-context path is unchanged from upstream, so this should reproduce on stock `main`.

## Symptom

For a ~44 s English dictation (16 kHz mono, peak ≈ 2.6 % FS — i.e. quiet), the transcript ends mid-sentence. The audio clearly continues for ~4.6 s after the last emitted token, with speech-level energy throughout (tail RMS ≈ the utterance's own speech RMS). `result.confidence` is ~0.98 and every emitted token is high-confidence — nothing signals the loss.

## Root cause (per-frame decoder trace)

Instrumenting `TdtDecoderV3.decodeWithTimings` to log the joint decision at each frame shows the utterance split into 4 windows (stride ≈ 161 frames). The first three decode normally. The **final** window (`isLastChunk = true`, `globalFrameOffset = 483`, 77 valid frames ≈ 6.1 s of real speech) predicts **blank (id 8192) at every frame**:

```
ENTER isLast=true encSeq=77 effSeq=76 actualFrames=76 globalOff=483 t0=1
main   t=1  tok=8192 p=0.924 dur=1 blank=true
blank  t=8  tok=8192 p=1.000 dur=4 blank=true
blank  t=12 tok=8192 p=1.000 dur=4 blank=true
blank  t=16 tok=8192 p=1.000 dur=4 blank=true
  ... (every frame to t=76) ...
blank  t=72 tok=8192 p=0.912 dur=4 blank=true
MAINLOOP-EXIT t=76 effSeq=76 emitted=0 lastTok=nil
FINALIZE begin t=76   -> recovers nothing
```

Key point: the decode loop does **not** terminate early — it walks all 76 frames and the joint calls every one silence at p≈1.0. So the finalization pass (`isLastChunk`) has nothing to flush; the encoder features for this window are already blank-like. Notably, the audio at the **start** of this window (its overlap region) was transcribed correctly by the *previous* window — the same samples read as speech there and as blank here. The only difference is that here they sit at the head of a short, heavily zero-padded 15 s encoder window with the mel-context prepend.

This looks like a new manifestation of the mel-context sensitivity already noted for #594 — but here it blanks the *whole* short final window rather than just drifting the first frames.

## What recovers it (three independent signals)

1. **Isolated tail** — feeding just the final ~6 s as its own clip (single-window path, no prepend) transcribes it perfectly, including a word the full run dropped.
2. **`melChunkContext = false`** — the no-mel/warmup path recovers the tail. (On a set of clean long-form clips it introduced no new tail drops.)
3. **Amplifying the input** — a moderate gain boost recovers the full transcript. (Not a robust fix on its own: recovery is non-monotonic in gain for some clips, and utterances with a loud word + quiet tail don't trigger a global-peak normalizer.)

## Reproduction

The drop needs the real interaction of low gain + pause structure + final-window alignment, so a synthetic "just attenuate a clip" recipe is unreliable (mild attenuation → word errors; heavy → the whole utterance collapses to empty). It reproduces consistently on genuine quiet, long-form recordings. **I can share a minimal anonymized sample WAV on request** (the ones I have are user dictation I can't post publicly).

Conditions that trigger it:
- `melChunkContext = true` (default), v3.
- Audio long enough for ≥3 windows (roughly >30 s).
- Low input gain (peak on the order of a few % FS).
- A pause and/or the dropped span landing in a short final window.

## Suggested direction

A targeted mitigation that's worked well for us: after the decode loop, if the final window emitted no tokens (or stopped well before the end of its own audio) **and** its untranscribed tail still carries speech-level energy, re-decode that window once with `contextSamples = 0` and substitute only if it reaches further — the rescued tokens then flow through the existing overlap merge/dedup unchanged. It's ~90 lines in `ChunkProcessor`, additive, fires only on the failure signature, and costs one extra decode pass when it does. Happy to open it as a PR if useful — though a root-cause fix in how mel-context is applied to a short final window may well be cleaner than a re-decode, and you'll know that path better than I do.

Two things I'd want your read on either way: the heuristic thresholds (tail-gap and the speech-energy ratio), and behavior on very marginal tails where the re-decode can produce a plausible-but-uncertain clause.

Alex-Wengg:
Closes #747.

## Problem

On quiet long-form audio the transcript ends several words early with high confidence and no error. Validated against the reporter's reproducer (17.28 s, 16 kHz mono, peak ≈ 1.7% FS — genuinely quiet). Stock `main` drops the final clause:

> "…phases without me." ❌ (missing: "as long as your confidence is reasonable.")

Root cause: the **final** sliding window is a short chunk zero-padded up to the model window, so the encoder sees a frame distribution dominated by digital silence. On speech that is already near the noise floor, that degenerate input pushes every emission to blank. The seam-gap repair (#761) only walks gaps *between* tokens, so a tail dropped past the last emitted token was never examined.

## Fix

**1. End-align the final window (structural).** A short final chunk fills its window backwards with real audio instead of zeros (`ChunkProcessor.lastChunkWarmupSamples`). The backfilled prefix rides the existing warmup mechanics — decoded from frame 0, tokens before the original chunk-start frame suppressed via `emitTokensAfterGlobalFrame` — so emitted coverage and the merge overlap with the previous chunk are unchanged. The model simply never sees a mostly-zeros window, and the final chunk gains real left acoustic context.

The window ends at the **last speech-bearing frame** (`speechEndSamples()`), not EOF: recordings often end in operator silence or digital zeros, and a window ending inside a dead-silence run *within its declared audio length* decodes degenerately even when the silence is recorded (measured on an Earnings-22 call: 12 s of speech decodes perfectly alone, loses its first half with 1 s of the recording's silent tail appended, decodes to **empty** with all 3 s — while zero padding beyond `audio_length` is masked and harmless). The trim threshold is `speechRmsFloor` — below any real speech — so nothing transcribable is excluded; an all-sub-floor tail is skipped outright.

No repair pass, no probe, no splice: an earlier iteration of this PR shipped a trailing-drop repair pass (probe + tail re-decode + splice + stale-punctuation trim); the structural fix makes the disease impossible instead of treating the symptom, and deletes all of it.

Prefix suppression is a V3-decoder feature: the tdtJa dispatch now forwards it (`TdtDecoderV3` underneath); v2/tdtCtc110m keep the zero-padded layout (`supportsSuppressedPrefix`) since `TdtDecoderV2` would emit the prefix twice. Applies to both the default loop and the dual-decode arbitration path. Files shorter than one window are unaffected (whole-file decode; padding there is unavoidable and harmless).

**2. Adaptive speech gate for the seam-gap repair.** The reproducer exposed a dead gate in the shipped seam-gap pass (#758/#761): its speech-energy probe used an absolute per-frame RMS threshold (0.008), but quiet recordings sit far below it — the gate could never fire on the audio class it guarded. Replaced with a threshold adapted to the recording's own level:

```
min(0.008, max(0.0005, p75FrameRms × 0.3))
```

Normal-level audio clamps to the previous 0.008 ceiling (existing behavior preserved); quiet recordings scale down with an absolute floor so dither never counts as speech. Frames of exact digital silence are excluded from the percentile (they are *no recording* and would collapse the gate to its floor on silence-heavy files). Constants are named (`speechRms*`) with a full value-rationale table in `LongTranscription.md`.

## Result

Fix branch output matches the reporter's expected transcript exactly:

> "Please do keep going autonomously. I won't be around much, so feel free to continue working through multiple steps or phases without me as long as your confidence is reasonable." ✅

## Verification

- Reproducer: byte-exact match with expected transcript (default `ASRConfig`, batch `transcribe`)
- 20-file LibriSpeech test-clean identical to stock `main` (0.7% avg WER, 0.2% avg CER) — all single-chunk, so additionally:
- **A/B vs stock `main` on 7 long multi-chunk files** — 4 hour-long Earnings-22 calls (WER-scored against references) + 3 long MUSAN recordings (7–10 min): **byte-identical on 5 of 7**; the only diffs are two ground-truth-verified tail improvements — "this connect" → "disconnect" (reference: "YOU MAY NOW DISCONNECT") and filler smoothing ("uh"/"Um and" → "And") on spontaneous speech. Per-file Earnings WER matches `main` to the last digit (13.14 / 20.17 / 9.97 / 26.71). The Earnings A/B also caught and fixed an intermediate regression during development — an EOF-aligned window embedding the recording's own trailing zeros — which is what motivated the speech-end trim.
- Unit tests: `EndAlignedFinalWindowTests` (backfill arithmetic: fill, caps by preceding audio, frame alignment, single-chunk no-op, non-final passthrough, trailing-silence trim, speech-end scan) + adaptive-threshold and digital-silence cases in `SeamGapRepairTests`
- swift-format lint clean; all CI checks green including the ASR Benchmark job

## Known limitations

- The dead-silence-at-window-end pathology also afflicts **mid-file** windows whose fixed-stride end lands inside a silence run (pre-existing on `main`; can drop nearby quiet speech — observed on a LibriVox recording's quiet outro credit). Snapping *every* window end to the last speech-bearing frame was tried and reverted: it perturbs dozens of mid-file windows per hour of audio for a net-neutral WER change. Documented in `LongTranscription.md` as follow-up-issue material.
- The adaptive gate's reference level is whole-file: a loud-body recording with a quiet mid-file gap clamps to the ceiling, so that gap is gated as if the whole file were loud. A gap-local reference would close this (seam-gap pass only; the trailing case no longer depends on the gate).
- v2/tdtCtc110m models keep the old zero-padded final window (their decoder lacks prefix suppression).

## Also included

- `Documentation/ASR/LongTranscription.md`: new "End-Aligned Final Window" section (including the measured silence-poisoning series), adaptive-gate constants table (value + dBFS + why-this-value per constant), updated invariants list and milestone table, and the #706/#708 and #759 merge fixes documented alongside the seam-gap pass.
- `ChunkProcessor` / `ASRConfig` comment slim-down: multi-paragraph rationale moved into `LongTranscription.md`; code keeps one-to-three-line pointers. Details that existed only in code (lowercase-canonical rule, edge-dedupe tolerance, probe-budget sizing, arbitration tie rule and overhead) now live in the doc.


jonyoder:
Confirmed fixed on my reproducer clip.

PR #802, opened 2026-07-16
fix(offline-diarizer): pyannote-parity clustering — threshold semantics, constraint count, constrained assignment

### LINKED ISSUE #801: Offline diarizer: auto clustering and numSpeakers=2 return the same speaker count but different partitions (69% vs 92% correctly-attributed speech)
## Summary

On a real 2-speaker conversation, `OfflineDiarizerManager` with `.default` config returns the **correct speaker count (2)** but a **wrong partition**. Pinning `clustering.numSpeakers = 2` returns the *same count* with a **materially different and correct partition**.

Same N, different answer. That asymmetry is the core of this report: if auto already estimates 2, I would expect it to then partition as if N=2.

Scored against a pyannote `speaker-diarization-community-1` reference on **identical bytes** (10ms frames, optimal label permutation, overlap-forgiven):

| config | speakers | speech correctly attributed | confused |
|---|---|---|---|
| `.default` (auto) | 2 | **69.0%** | 25.07s |
| `clustering.numSpeakers = 2` | 2 | **92.0%** | 6.09s |

## Detail

Audio: 86.5s, 16kHz mono, two speakers, a balanced 63/37 split by speech time. Speaker B holds `0.0-7.0` and then `62.0-86.9`; speaker A holds the middle.

- **auto** partitions at **82.0s**: `A 0.0-82.0`, `B 82.0-86.5`. Speaker B's 25-second closing stretch is absorbed into A.
- **`numSpeakers=2`** partitions at **62.0s**: `0-62` / `62-86.5`, which matches the reference except for the `0-7` opening turn.
- The `62.0` boundary is genuinely found by both engines independently, so segmentation is proposing it; the auto path just does not partition on it.
- Deterministic: 3 identical runs, byte-identical output.

## Other observations from a 20+ config sweep

1. **`Clustering.threshold` appears inverted relative to its docstring.** It is documented as a "Euclidean distance threshold for unit-normalized embeddings", so lowering it should split more. Observed: `0.45`, `0.4`, `0.3` all **merge to 1 speaker**, while `0.5`, `0.6` (stock), `0.7`, `0.9`, `1.1` are **byte-identical** to each other. Also `numSpeakers=2 + threshold=0.5` reverts to the bad 69% partition, i.e. setting threshold to a value that is otherwise a no-op cancels the `numSpeakers` win. Is the direction intended, and is it meant to interact with `numSpeakers` this way?
2. **`minSpeakers=2` alone is a no-op here** (byte-identical to stock), which makes sense since auto already returns 2, so the floor never binds. `minSpeakers=2, maxSpeakers=2` is byte-identical to `numSpeakers=2`. Worth a docs note: only an exact count (or min+max together) changes the partition; the floor alone cannot correct an under-partition.
3. **`numSpeakers` behaves as a target, not a constraint.** On a separate 15.8s clip, `numSpeakers=4` returned 3 and `numSpeakers=3` returned 2. That seems reasonable, but it is not obvious from the API surface, and it does mean callers cannot rely on the requested count. A docs note would help.
4. `ZeroVoteReembed(enabled: true)` at `minDurationSeconds` 0.4 / 0.25 / 0.1 is **byte-identical** to stock on this audio (with and without `numSpeakers`), so it does not address this case. Its doc comment describes a similar-sounding symptom ("silently absorbing whole speaker turns into the surrounding speaker's segment"), which is what led me to try it.
5. `Embedding.minSegmentDurationSeconds = 0.25` (vs stock 1.0) is also byte-identical here.

## Environment

- FluidAudio 0.15.5 (exact pin), models `FluidInference/speaker-diarization-coreml`.
- Reproduced on macOS 26.5 (arm64, Swift 6.3) via a small harness calling `OfflineDiarizerManager` directly, and observed in an app on iPadOS (iPad Pro M2).
- One caveat I want to be upfront about: on this clip the macOS harness reports 2 speakers where the iPad ANE reports 1. The device is *worse*. I have not explained that gap; it may be marginal numerical differences on a borderline cluster. On a 4-speaker control clip, harness and device agree exactly (4 vs 4).

## What I can and cannot share

The audio is a private personal recording, so I cannot attach it. I am happy to:
- run any config or patch you suggest against it and report numbers,
- try to reproduce on a public corpus if you can point me at a case with a similar shape (one speaker opening briefly, then returning for a long closing stretch).

I also want to flag, in case it saves someone time: I initially "confirmed" this was unfixable by scoring configs against a **truncated** reference (I had capped the reference turn list, which hid the closing 25 seconds). That produced a confident and completely wrong conclusion until I uncapped it. Given #752 turned out to be placeholder ground truth in the benchmark, corrupted references seem to be an easy trap in this problem space.

## Questions

1. Is auto-vs-`numSpeakers=N` producing different partitions at the same N expected behaviour, or a bug in the count-estimation / VBx warm-start path?
2. Is `threshold`'s observed direction (lower = fewer clusters) intended?

Alex-Wengg:

Fixes #801: `OfflineDiarizerManager` with `.default` config and with `clustering.numSpeakers = 2` returned the **same speaker count (2) but materially different partitions** (69% vs 92% correctly-attributed speech on the reporter's 2-speaker clip). Every observation in the reporter's 20-config sweep decodes exactly to three porting bugs against pyannote community-1's `VBxClustering`:

**1. AHC threshold was inverted (issue Q2 — confirmed).** The config value is documented as a "Euclidean distance threshold for unit-normalized embeddings", and pyannote applies it directly as the dendrogram cut: `fcluster(dendrogram, threshold, criterion="distance")`. Our port reinterpreted it as a cosine similarity and cut at `sqrt(2 − 2t)`, which inverts the knob (raising it split more instead of merging more) and shifted the default 0.6 to an effective cut of 0.894. The reporter's sweep maps perfectly through this formula: 0.3/0.4/0.45 → cuts 1.18/1.10/1.05 (merged to 1 speaker), while 0.6–1.1 → cuts 0.894–0.0 (byte-identical near-singleton AHC output). The value is now applied directly, clamped/validated to (0, 2].

**2. Speaker-count constraints compared against the wrong count (issue Q1 — it's a bug).** pyannote decides whether to fall back to K-Means by comparing the requested count against `auto_num_clusters` — the clusters VBx actually kept (`pi > 1e-7`). We compared against the AHC warm-start count, which on this audio was far above 2 (AHC produced near-singletons, VBx collapsed them to 2). So `numSpeakers=2` always triggered K-Means re-clustering — a completely different algorithm — even though auto had already detected 2. Hence same N, different partition. The check now uses the new `VBxOutput.activeClusterCount`. This also explains the reporter's side observations: `minSpeakers=2` never bound (compared against the inflated count), and `numSpeakers=2 + threshold=0.5` "cancelling the win" (cut 1.0 made AHC itself return 2, so the constraint stopped binding).

**3. Missing constrained assignment (the actual quality gap).** Fixing (2) alone would regress the reporter's clip to the bad 69% partition — the K-Means path was accidentally winning. pyannote assigns embeddings with `constrained_argmax`: local speakers sharing a segmentation chunk must map to **distinct** clusters (Hungarian matching per chunk, maximizing total similarity). Our plain per-embedding argmax let two co-chunk speakers snap to the same centroid, silently absorbing one speaker's turns into another's — the mechanism behind the bad 82s-boundary partition. Ported it behind `Clustering.constrainedAssignment` (default true), auto-disabled when the count is forced via K-Means, matching pyannote. The Hungarian solver moved from `DiarizationDER.swift` into an internal `HungarianAssignment` enum with a rectangular max-score wrapper.

### Behavior change / migration

The `threshold` knob keeps its numeric default (0.6, the community-1 value) but its **meaning changes to pyannote's**: larger = more merging = fewer speakers. Old values map to new ones via `sqrt(2 − 2·old)` — old 0.6 behaved like 0.894, and the old AMI advice of `--threshold 0.7` corresponds to `0.775` today. `Benchmarks.md` and CLI help are updated; the AMI-SDM offline table should be re-benchmarked on this branch since it was measured under the old semantics.

### Validation

- Unit tests for the threshold direction/extremes, `activeClusterCount`, the Hungarian solver, and the constrained assigner (including the exact #801 failure mode: two co-chunk embeddings both nearest to the same centroid).
- `swift build` + swift-format clean; XCTest runs in CI (unavailable locally).
- The reporter offered to re-run patches against their private audio — a build of this branch with stock `.default` config is the ideal check; the prediction is the auto path now finds the 62s boundary on its own.
- Caveat: pyannote community-1's `config.yaml` is gated on HF, so the shipped `ahc_threshold` (assumed 0.6-as-distance, within the code's `Uniform(0.5, 0.8)` prior) is inferred from source semantics, not read from the config.

### DIFF
--- a/Documentation/Benchmarks.md
+++ b/Documentation/Benchmarks.md
@@ -634,7 +634,7 @@ AVERAGE          10.6     17.4       5.4      2.0      3.3      -        323.2
 
 12/16 meetings detect the correct speaker count. Average DER 10.62% matches published pyannote-community-1 offline numbers on this split (~11-12%). Results are fully deterministic (two consecutive runs produce bit-identical metrics).
 
-**Clustering threshold matters on this split.** The table above uses `--threshold 0.7`. Since #616 the CLI default is the community-1 preset (0.6), which merges clusters more aggressively and undercounts speakers on 4 meetings (EN2002a 2/4 → 41.9% DER, ES2004d 3/4 → 34.8%, EN2002b 19.0%, IS1009d 16.4%), degrading the average to 15.5% DER. Pass `--threshold 0.7` (or set `clusteringThreshold: 0.7` in `OfflineDiarizerConfig`) for AMI-SDM-like meeting audio.
+**Clustering threshold matters on this split.** The table above predates #801 and uses `--threshold 0.7` under the old (inverted) threshold semantics. Since #801 the threshold is a Euclidean cut distance applied directly to the AHC dendrogram (pyannote parity): **larger values merge more aggressively and yield fewer speakers**. Old values map to new ones via `sqrt(2 − 2·old)` — the old default 0.6 behaved like a cut at 0.894, and the old `--threshold 0.7` from this table behaves like `--threshold 0.775` today. The CLI default remains the community-1 preset value (0.6), now interpreted as pyannote does.
 
 ### Streaming/online Diarization
 
--- a/Sources/FluidAudio/Diarizer/DiarizationDER.swift
+++ b/Sources/FluidAudio/Diarizer/DiarizationDER.swift
@@ -121,7 +121,7 @@ public enum DiarizationDER {
                     cost[h * n + r] = maxO - overlap[h * R + r]
                 }
             }
-            let assign = Hungarian.solve(costSquare: cost, n: n)
+            let assign = HungarianAssignment.solve(costSquare: cost, n: n)
             for h in 0..<H {
                 let r = assign[h]
                 if r < R && overlap[h * R + r] > 0 {
@@ -227,66 +227,5 @@ public enum DiarizationDER {
             }
         }
         return mask
-    }
-}
-
-// MARK: - Hungarian (O(n^3) min-cost assignment on a square cost matrix)
-
-fileprivate enum Hungarian {
-    /// Kuhn-Munkres with potentials. `cost` is row-major `n × n`,
-    /// non-negative integers. Returns `assign[row] = col`.
-    static func solve(costSquare cost: [Int], n: Int) -> [Int] {
-        if n == 0 { return [] }
-        // Classic Jonker-Volgenant style implementation adapted for
-        // square n×n. 1-based indexing in arrays of size n+1 for the
-        // canonical algorithm — cost is read as `cost[(i-1)*n + (j-1)]`.
-        let INF = Int.max / 4
-        var u = [Int](repeating: 0, count: n + 1)
-        var v = [Int](repeating: 0, count: n + 1)
-        var p = [Int](repeating: 0, count: n + 1)
-        var way = [Int](repeating: 0, count: n + 1)
-
-        for i in 1...n {
-            p[0] = i
-            var j0 = 0
-            var minv = [Int](repeating: INF, count: n + 1)
-            var used = [Bool](repeating: false, count: n + 1)
-            repeat {
-                used[j0] = true
-                let i0 = p[j0]
-                var delta = INF
-                var j1 = 0
-                for j in 1...n where !used[j] {
-                    let cur = cost[(i0 - 1) * n + (j - 1)] - u[i0] - v[j]
-                    if cur < minv[j] {
-                        minv[j] = cur
-                        way[j] = j0
-                    }
-                    if minv[j] < delta {
-                        delta = minv[j]
-                        j1 = j
-                    }
-                }
-                for j in 0...n {
-                    if used[j] {
-                        u[p[j]] += delta
-                        v[j] -= delta
-                    } else {
-                        minv[j] -= delta
-                    }
-                }
-                j0 = j1
-            } while p[j0] != 0
-            repeat {
-                let j1 = way[j0]
-                p[j0] = p[j1]
-                j0 = j1
-            } while j0 != 0
-        }
-        var assign = [Int](repeating: -1, count: n)
-        for j in 1...n {
-            if p[j] != 0 { assign[p[j] - 1] = j - 1 }
-        }
-        return assign
     }
 }
--- a/Sources/FluidAudio/Diarizer/HungarianAssignment.swift
+++ b/Sources/FluidAudio/Diarizer/HungarianAssignment.swift
@@ -1,0 +1,98 @@
+import Foundation
+
+// MARK: - Hungarian (O(n^3) min-cost assignment on a square cost matrix)
+
+enum HungarianAssignment {
+    /// Kuhn-Munkres with potentials. `cost` is row-major `n × n`,
+    /// non-negative integers. Returns `assign[row] = col`.
+    static func solve(costSquare cost: [Int], n: Int) -> [Int] {
+        if n == 0 { return [] }
+        // Classic Jonker-Volgenant style implementation adapted for
+        // square n×n. 1-based indexing in arrays of size n+1 for the
+        // canonical algorithm — cost is read as `cost[(i-1)*n + (j-1)]`.
+        let INF = Int.max / 4
+        var u = [Int](repeating: 0, count: n + 1)
+        var v = [Int](repeating: 0, count: n + 1)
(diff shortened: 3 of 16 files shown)


sunhyung:
Not the original reporter, but I have a case that stresses this harder than the
86.5s clip in #801, so here are numbers on both 0.15.5 and this branch.

**Setup.** Offline pipeline via `OfflineDiarizerManager`, models
`FluidInference/speaker-diarization-coreml`, Apple Silicon / macOS 26.
Audio is a 63.7-minute, 16 kHz mono meeting with **7 speakers** (count known
first-hand, not estimated). It is a private recording so I cannot attach it, but
I am happy to run any config or patch against it.

Branch tested at `1b0e544`.

### The collapse this fixes

On 0.15.5 the stock config puts the entire meeting on one speaker — segmentation
proposes 458 spans and clustering merges every one of them:

| config (0.15.5) | speakers | spans |
|---|---|---|
| stock | **1** | 458 |
| `threshold` 0.20 … 0.65 (8 steps) | 1 | 458 (byte-identical) |
| `threshold` 0.68 | 3 | — |
| `threshold` 0.72 / 0.75 / 0.78 | 6 | — |
| `threshold` 0.85 | 5 | — |
| `threshold` 0.90 / 1.0 / 1.2 / 1.4 | 4 | — |

On this branch, stock config alone gets most of the way there:

| config (#802 @ 1b0e544) | speakers | spans |
|---|---|---|
| stock | **6** | 519 |
| `numSpeakers = 7` | 7 | 535 |
| `maxSpeakers = 3` | 3 | 513 |
| `minSpeakers = 8` | 8 | 546 |

1 → 6 out of a true 7, with no tuning. That is the headline for me.

### The old→new threshold mapping checks out

Your PR says old 0.6 behaved like an effective cut of 0.894. On this branch
`threshold = 0.894` reproduces the old stock result exactly — 1 speaker — which
is a nice independent confirmation of `sqrt(2 − 2t)` on unrelated audio.

### Constraints now bind (they did not before)

On 0.15.5, `minSpeakers` was inert in the direction that mattered: with stock
detecting 1, `minSpeakers` of 2, 3 and 5 each returned 1 speaker / 458 spans,
byte-identical to passing nothing. `maxSpeakers` behaved as a target rather than
a ceiling (10 → exactly 10).

Both are fixed here: `minSpeakers = 8` returns 8, and `maxSpeakers = 3` returns 3.
Note this is a stronger case than #801's item 2 — there the floor never had to
bind because auto already returned the requested count. Here auto returns 1 and
the floor still did nothing, so it was a real failure to enforce, and your
diagnosis (comparing against the AHC warm-start count instead of the VBx active
count) predicts exactly that.

### One thing that still looks inconsistent

On a 60-second excerpt of the same meeting, `numSpeakers = 5` is a no-op, and the
reason is a mismatch between the count the constraint sees and the count the
pipeline reports. I printed `detectedCount` at the top of
`refineWithConstraints`:

```
[[PROBE]] detected=5 min=5 max=5 num=Optional(5) needs=false   → pipeline reports 3 speakers
[[PROBE]] detected=5 min=2 max=2 num=Optional(2) needs=true    → pipeline reports 2 speakers
```

So `activeClusterCount` is 5 while the user-visible result is 3: two clusters
survive the `pi > epsilon` test but end up with no embeddings assigned to them.
Asking for 5 is therefore silently ignored, because the constraint believes it
already has 5.

Is `activeClusterCount` meant to be the count callers see? If the reported count
comes from `centroids.count` after assignment, the constraint check and the
output can disagree — which feels like the same class of mismatch this PR is
fixing, just moved one step later in the pipeline.

### Caveat on my own numbers

Worth flagging since #801 already has one of these: my first pass at the
constraint table showed *every* constraint as a no-op on this branch, and I was
about to report it as a regression. It was my harness — zsh does not word-split
unquoted parameter expansions, so `--speakers 2` was arriving as a single
argument and never reaching the config. The tables above are from re-runs with
that fixed, and the probe output above is what caught it.

Alex-Wengg:
@sunhyung  i added a fix lmk if it works on your end

Issue #702, opened 2026-06-15
CTC custom-vocabulary rescorer over-fires on short keyword terms since 0.14.8 (PR #634)


jonyoder:
### Summary

PR #634 ("Fix/word boost improvements", first in **0.14.8**) reworked the CTC vocabulary-rescoring pipeline. It clearly improved KWS on the validation sets it targeted (earnings22 company names, FDA drug names — long, distinctive terms). But for **short keyword terms that collide acoustically with common English** (e.g. `CRAN`~"ran", `Snyk`~"sync", `Redis`~"ready", a 3-letter acronym ~ a function word), it now **over-fires badly**: ordinary words in the transcript get replaced by vocabulary terms.

### Reproduce

Custom vocabulary of short terms; transcribe ordinary speech that contains **none** of them but plenty of near-homophones. Example with `cbw=2.0, minSimilarity=0.65`:

- "The overnight job **ran** twice…" → "The overnight job **CRAN** twice…"
- "Rebuild the search **index**…" → "Rebuild the search **bindex**…"
- "The sensor alerts the team at 50 **parts per million**." → "**RFD** … **PPM** **CRAN** **RFD**" (whole sentence shredded)

### Bisect

Clean on **0.14.5**; over-fires on **0.14.8** and all **0.15.x**. `410044d1` is the only `CustomVocabulary/` commit in `0.14.5..0.14.8`.

### This isn't tunable from the public API

On 0.15.3 I swept `minSimilarity` (0.65–0.92), `marginSeconds` (0.10–0.50), `cbw` (0.5–2.0), and `TDT_EMISSION_DELAY_FRAMES` (0/1/3). A 54-clip ordinary-speech corpus (3 voices) with an 8-term short-vocab dictionary:

|                                   | false-positive insertions | recall on a true-positive set |
| --------------------------------- | ------------------------- | ----------------------------- |
| **0.14.5**                        | **12**                    | 98.3%                         |
| 0.15.3, best of *any* knob combo  | **~54**                   | ≤95%                          |

Nothing approaches 0.14.5; tightening the gates only destroys recall. **The diagnostic tell: the false matches now score *higher* similarity than the true matches**, so no `minSimilarity` threshold separates them — which points at the blank-aware DP score scale rather than a tuning default.

### Asks (any one would unblock us)

1. An option to select the **pre-#634 scoring path** for short-vocab KWS, or
[...]

PR #805, opened 2026-07-18
feat(asr): expose non-mutating vocabulary candidate evidence

CiprianSpiridon:

VocabularyRescorer currently exposes only RescoreOutput after legacy vocabulary replacements have already been applied. Applications that need a different precision and recall policy cannot preserve the decoder transcript while inspecting the accepted and rejected CTC comparisons that produced those replacements.

PR #722 added useful opt-in safety controls, but a global threshold still cannot express context-aware arbitration for collision-prone technical terms without sacrificing accented and domain-term recall. This additive API lets callers retain the untouched transcript and make their own final selection from FluidAudio evidence without changing existing scoring defaults or legacy output.

This is an additive follow-up to #702 and #722; it does not claim to change the default false-positive behavior.

### Summary

- Add public Sendable candidate-evidence and output types.
- Add ctcTokenEvaluateCandidates, which returns the untouched base transcript plus every candidate that reaches CTC evaluation, including legacy rejections.
- Report canonical term, exact matched alias, string similarity, raw vocabulary and original CTC scores, effective boost, half-open word and token ranges, timestamps, legacy decision, and reason.
- Preserve canonical-versus-alias provenance and contiguous token provenance across word-centric, term-centric, and spotter-rescue paths.
- Keep ctcTokenRescore source-compatible and avoid allocating an evidence buffer on the legacy path.
- Document the non-mutating API and add focused unit coverage for score mapping, alias identity, unavailable values, decisions, and span semantics.

### Compatibility

- No existing public API is removed or renamed.
- No scoring threshold, model, configuration default, or legacy replacement rule changes.
- Existing ctcTokenRescore callers continue to receive the same rewritten output contract.

### Test Plan

- [x] swift format lint on every changed Swift file
- [x] swift build
- [x] swift test --parallel --num-workers 8: 2,116 XCTest cases and 45 Swift Testing cases passed
- [x] VocabularyCandidateEvidenceTests: 6 of 6 passed
- [x] xcodebuild -scheme FluidAudio -destination generic/platform=iOS build
- [x] Full branch diff reviewed against current canonical main with no blocking findings

### DIFF
--- a/Documentation/ASR/CustomVocabulary.md
+++ b/Documentation/ASR/CustomVocabulary.md
@@ -306,25 +306,64 @@ let output = rescorer.ctcTokenEvaluateCandidates(
 
 // Always identical to the supplied decoder transcript.
 let untouchedText = output.baseText
+print(output.baseWords) // Exact source-word sequence indexed by every wordRange.
 
 for candidate in output.candidates {
-    print(candidate.basePhrase, candidate.canonicalTerm)
+    print(candidate.candidateID, candidate.origin)
+    print(candidate.basePhrase, candidate.canonicalTerm, candidate.matchedAlias as Any)
     print(candidate.rawOriginalCTCScore, candidate.rawVocabularyCTCScore)
-    print(candidate.effectiveBoost, candidate.legacyDecision)
+    print(candidate.effectiveBoost as Any, candidate.comparisonPassed)
+    print(candidate.legacyOutcome, candidate.wordRange, candidate.tokenRange as Any)
+    print(candidate.baseTextUTF8Range as Any)
 }
 ```
 
 The candidate API runs the same discovery, guards, and CTC comparison as legacy rescoring, but it
-returns both accepted and rejected evaluations without returning a rewritten transcript. Each
-candidate identifies the canonical term, the exact alias that produced the best string match (or
-`nil` for the canonical form), string similarity, raw CTC scores before context biasing, the
-effective boost, and FluidAudio's legacy accept/reject decision and reason.
+returns every comparison evaluation without rewriting `baseText`. Each candidate identifies the
+canonical term, the exact alias that produced the best string match (or `nil` for the canonical
+form), string similarity, raw CTC scores before context biasing, and the effective boost.
 
-`wordRange` and `tokenRange` are half-open ranges into the base transcript's word sequence and the
-supplied `tokenTimings`, respectively. `tokenRange` is `nil` when the source tokens are not
-contiguous. Scores and timestamps are optional rather than sentinel values when the corresponding
-evidence is unavailable. Candidate order is diagnostic; callers should use the reported ranges and
-scores rather than array position to arbitrate overlapping proposals.
+`comparisonPassed` reports only the numeric, pre-arbitration comparison: boosted vocabulary score
+greater than original score. `legacyOutcome` reports what the compatibility `ctcTokenRescore()`
+path finally did after overlap arbitration:
+
+- `applied` — comparison passed and the candidate was written to the legacy transcript.
+- `rejectedByComparison` — the legacy floating-point comparison ran and rejected the candidate.
+- `unavailableEvidence` — the comparison could not run, for example because tokenizer evidence was
+  unavailable.
+- `supersededByOverlap` — comparison passed, but an overlapping candidate won final arbitration.
+
+`comparisonPassed` and `legacyOutcome` preserve the comparison FluidAudio actually performed. A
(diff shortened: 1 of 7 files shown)


Alex-Wengg:
some code review from claude bot. 


1. legacyDecision is recorded before arbitration. Evidence is appended right after evaluateCTCMatch, but finalizeReplacements' greedy overlap pass runs afterward and can drop candidates marked .accepted. So for the overlap case documented in this very file (rom vimza→Romvimza vs vimza→Cimzia), both come back .accepted while legacy applies only one. A caller replaying all .accepted candidates diverges from FluidAudio's own output, and the docs say the field is "the decision made by the existing replacement behavior" — which it isn't. Suggest either a wasApplied (post-arbitration) field alongside the raw comparison, or renaming/re-documenting the field as pre-arbitration.
2. wordRange indexes a word list the caller can't see. The ranges index the internal buildWordTimings(from: tokenTimings) sequence (skips <blank>/<pad>/empty tokens, drops empty-trimmed words), not a whitespace split of baseText. The docs say "the base transcript's word sequence," which invites exactly the wrong indexing. Either return the word array (or string-index spans into baseText), or document precisely what the ranges refer to and the transcript/tokenTimings alignment requirement.
3. The reason string lies on rejections. It's built unconditionally with a hard-coded > ("'Term'=-8.00 > 'orig'=-3.00") independent of shouldReplace, so rejected evidence carries text contradicting its own scores. Pre-existing string, but the old code only surfaced it on accepts — the evidence API newly exposes it on the reject path.
4. Duplicate rows are unattributable. The span sweep evaluates every qualifying span at the same start (length 2 and 3 can both land, one accepted + one rejected for the same term/region), and the spotter-rescue pass can add more (its seenSpans dedupe is local). There's no field identifying the producing pass, so keying evidence by term or time window is ambiguous.

CiprianSpiridon:
Thanks — all four findings were correct and substantive. I pushed follow-up commit `b103f159` after reviewing the complete PR diff end to end.

1. **Pre-comparison vs. final application are now separate.** `comparisonPassed` records only the acoustic comparison. `LegacyApplicationOutcome` records the terminal compatibility result: `applied`, `rejectedByComparison`, `unavailableEvidence`, or `supersededByOverlap`. Final outcomes are reconciled after the same greedy overlap arbitration used by `ctcTokenRescore()`. The Romvimza/Cimzia overlap regression now asserts that both comparisons can pass while only one is `applied` and the other is `supersededByOverlap`.

2. **`wordRange` now has authoritative provenance.** `CandidateEvidenceOutput` returns the exact internal `baseWords` sequence that every `wordRange` indexes. Evidence also exposes a contiguous `tokenRange` when available and an optional half-open `baseTextUTF8Range`. Text ranges are emitted only after a unique, byte-exact alignment of the complete internal word sequence; punctuation, contractions, hyphens, Unicode, blank/pad gaps, repeated words, and ambiguous syntax are covered. Missing or ambiguous provenance returns `nil` instead of guessing.

3. **Comparison reasons are truthful.** Passed numeric comparisons use `>` and rejected comparisons use `<=`. Skipped comparisons do not invent a numeric reason. Raw scores and boosts are optional rather than sentinels; if FluidAudio completed a floating-point comparison with a non-finite diagnostic, the actual comparison/final outcome remains authoritative while the unavailable raw value stays `nil` and the display reason omits non-finite score text.

4. **Every discovery path is attributable.** Evidence now includes `CandidateOrigin` (`wordCentric`, `termCentricSingleWord`, `termCentricMultiWord`, or `spotterRescue`) and a non-negative `candidateID` unique within that `CandidateEvidenceOutput`. Duplicate-looking evaluations remain separate rows with separate IDs/origins, and the ID links each evaluation to its final overlap outcome. The docs explicitly state that IDs are output-local and must not be compared across evaluations.

The follow-up also keeps evidence collection unallocated on the legacy API path and makes UTF-8 mutation provenance fail closed for unrelated boundary symbols such as emoji and trademarks while preserving technical tokens such as `C++`, `C#`, `.NET`, `@MainActor`, `#if`, and `$PATH`.

Validation on the pushed tree:

- changed-file `swift format lint`: clean
- `swift build`: passed
- complete `swift test --parallel --num-workers 8`: 2,133 XCTest cases plus 45 Swift Testing cases passed
- focused candidate evidence/arbitration: 23 passed
- focused custom vocabulary: 19 passed
- generic iOS `xcodebuild`: passed

Ready for re-review. Thanks again for catching the contract-integrity gaps before downstream callers relied on them.

PR #806, opened 2026-07-18
fix(tts): make LuxTTS resources signable on iOS

CiprianSpiridon:

The LuxTTS G2P resources are currently copied as a top-level directory named `Resources`. In iOS resource bundles, Apple code signing interprets that reserved layout as a malformed legacy bundle and fails with `bundle format unrecognized, invalid, or unsuitable`. This prevents any signed iOS consumer build from using current FluidAudio `main`.

This change lets SwiftPM process the production and test resources into their bundle roots and updates the corresponding lookups. The resource contents and LuxTTS behavior are unchanged. A manifest comment and regression test protect the signing-sensitive layout from being reverted to `.copy`.

### Validation

- Changed-file `swift format lint`: clean; repository-wide lint exits 0 with existing warnings
- Clean `swift test --filter LuxTts`: 33 tests executed, 1 model-gated test skipped, 0 failures
- Full `swift test --parallel --num-workers 8`: exited 0 across 2,111 XCTest cases and 45 Swift Testing cases
- Signed generic iOS `FluidAudio` scheme build: passed
- Generated iOS production and test resource bundles contain their files at the bundle root and no nested `Resources` directory
- `codesign --verify --strict` passed for both generated iOS resource bundles

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -29,7 +29,7 @@ let package = Package(
             ],
             path: "Sources/FluidAudio",
             resources: [
-                .copy("TTS/LuxTts/G2p/Resources")
+                .process("TTS/LuxTts/G2p/Resources")
             ]
         ),
         // Byte-exact NeMo text normalization (FST engine, all 7 languages).
--- a/Package.swift
+++ b/Package.swift
@@ -29,6 +29,7 @@ let package = Package(
             ],
             path: "Sources/FluidAudio",
             resources: [
+                // Keep .process: .copy of a Resources-named directory breaks Apple code signing on iOS.
                 .process("TTS/LuxTts/G2p/Resources")
             ]
         ),
@@ -66,7 +67,7 @@ let package = Package(
                 "FluidAudioCLI",
             ],
             resources: [
-                .copy("TTS/LuxTts/Resources")
+                .process("TTS/LuxTts/Resources")
             ]
         ),
     ],
--- a/Sources/FluidAudio/TTS/LuxTts/G2p/LuxTtsG2p.swift
+++ b/Sources/FluidAudio/TTS/LuxTts/G2p/LuxTtsG2p.swift
@@ -77,11 +77,9 @@ public struct LuxTtsG2p: Sendable {
     public init() throws {
         guard
             let lexiconURL = Bundle.module.url(
-                forResource: "luxtts_en_us_lexicon.tsv", withExtension: "zz",
-                subdirectory: "Resources"),
+                forResource: "luxtts_en_us_lexicon.tsv", withExtension: "zz"),
             let auxURL = Bundle.module.url(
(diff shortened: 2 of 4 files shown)


Alex-Wengg:
1. The test target has the same latent issue. Package.swift line 69 still declares .copy("TTS/LuxTts/Resources") with matching subdirectory: "Resources" lookups in Tests/FluidAudioTests/TTS/LuxTts/LuxTtsFixtures.swift. That's the exact reserved-Resources layout this PR removes from the library target, it would hit the same CodeSign rejection if the test bundle is ever signed for device, and more importantly it's the in-repo example a future contributor will copy when adding bundled resources. Could you switch it to .process + drop the subdirectory: argument (or rename the directory) in this PR?


3. Add a one-line comment at the .process declaration noting that .copy of a Resources-named directory breaks Apple code signing on iOS. The constraint is invisible to every check we run — swift build, swift test, and CI (macOS-only tests, unsigned iOS build) all pass with either spelling — so without a comment, a revert-to-.copy "cleanup" would only surface in a downstream consumer's signed build.

CiprianSpiridon:
Thanks — both findings were correct. I pushed follow-up commit `4ddcfbc`.

1. The `FluidAudioTests` target now uses `.process("TTS/LuxTts/Resources")`, and `LuxTtsFixtures` resolves fixtures from the bundle root with no `subdirectory: "Resources"` assumption.
2. The production `.process` declaration now has the requested one-line warning explaining that copying a directory named `Resources` breaks iOS code signing.

I also added `testFixtureResourcesAreProcessedAtBundleRoot` to pin the actual bundle contract. Before the manifest fix, it failed because the generated test bundle contained a nested `Resources` directory and no root-level fixture. From a clean build after the fix, it passes and verifies both the absence of that directory and the root-level fixture.

Validation on the pushed head:

- changed-file `swift format lint`: clean; repository-wide lint exits 0 with existing warnings
- clean LuxTTS suite: 33 tests executed, 1 explicitly model-gated test skipped, 0 failures
- full `swift test --parallel --num-workers 8`: exited 0 across 2,111 XCTest cases and 45 Swift Testing cases
- signed generic iOS `FluidAudio` scheme build: passed
- generated iOS production and test resource bundles are flat and signed by Team `672Q5SQPSB`
- `codesign --verify --strict` passed for both bundles

For completeness, the package-wide generic-iOS `build-for-testing` signs the fixed test resource bundle successfully, then the overall `FluidAudio-Package` scheme stops on the existing intentional `#error("FluidAudioCLI is only supported on macOS")` because that aggregate scheme also includes the CLI. The supported `FluidAudio` iOS scheme is green.

Ready for re-review. Thanks for catching the mirrored test-target layout and the missing maintenance guard.

PR #824, opened 2026-07-29
Add ModelRegistry.repoOverrides so mirrored repos can be resolved

omachala:
`baseURL` already lets you point FluidAudio at a different registry host. What it does not let you do is point at a different *owner*, because the repository paths come from the `Repo` enum and those are baked in as `FluidInference/...`.

That gap shows up as soon as you mirror the models. We ship FluidAudio in an iOS app, and we want the model downloads our users trigger to come from a namespace we control, so we can read the release notes and test a new build before it reaches anyone. Setting `baseURL` to our own Hugging Face org does not get us there, since the path still says `FluidInference/`. Editing the enum does, but it has 82 cases and you add to it regularly, so we would be rebasing on every release.

A map felt like the smaller change:

```swift
ModelRegistry.repoOverrides = [
    "FluidInference/parakeet-tdt-0.6b-v3-coreml": "DictionLabs/parakeet-tdt-0.6b-v3-coreml"
]
```

Applied in both places that build a repo URL, `apiModels` and `resolveModel`, so listing and downloading can never point at different repos. Matching is longest-prefix, which picks up the cases carrying a subpath like `parakeet-realtime-eou-120m-coreml/160ms`. Anything unmapped resolves upstream unchanged, so a partial mirror is safe and new models you add keep working with nobody touching their config.

The default is an empty dictionary, so behavior is byte-identical unless someone sets it. It follows the same `nonisolated(unsafe)` static pattern as `_customBaseURL` directly above it.

Tests cover the no-op case, redirect through both call sites, fall-through for unmapped repos, subpath mapping, longest-prefix precedence, and composition with a custom `baseURL`.

Happy to rename it or change the matching rule if you would rather it behaved differently. We are running this in a fork either way, so there is no pressure on your side.

Disclosure: the code and this description were written in cooperation with AI coding agents.

### DIFF
--- a/Sources/FluidAudio/ModelRegistry.swift
+++ b/Sources/FluidAudio/ModelRegistry.swift
@@ -41,11 +41,49 @@ public enum ModelRegistry {
         }
     }
 
+    // MARK: - Repository Overrides
+
+    // Mutable static for runtime configuration, matching the `_customBaseURL` pattern above.
+    nonisolated(unsafe) private static var _repoOverrides: [String: String] = [:]
+
+    /// Maps upstream repository paths to alternates, e.g. a mirror under a different owner.
+    ///
+    /// `baseURL` only repoints the *host*, but repository paths come from the hardcoded `Repo`
+    /// enum, so a mirror hosted under a different owner on the same registry is unreachable
+    /// without this. Matching is longest-prefix, so mapping a repo also maps its subpaths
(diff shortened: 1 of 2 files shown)


omachala:
Hi @BrandonWeng 👋 just to let you know, I am around if you have any questions or any feedback..

Issue #855, opened 2026-08-16
Streaming: final flush window drops trailing words (mid-window entry, then carried-state re-decode)


saurabhav88:
## Summary

`duplicateFrameTolerance` (introduced by #788 / `667181a3`) makes sliding-window token dedup temporally
gated. On genuinely repetitive speech that gate changes `--streaming` output in two ways. Both reproduce on
stock `main`, both are isolated to that one constant by a one-line mutation, and both are regressions
against this repo's own batch decode of the same audio.

1. **Trailing words dropped.** Synthetic clip attached, so this one is fully reproducible from this issue.
2. **Phantom fragment inserted at a chunk seam.** Real dictation; described but not attachable.

We were about to ship #788 and pulled it back on this evidence. Not asking for a revert — happy to test a
patch.

## Environment

- `main` @ `c7b13a3942e79893f3bd76bfe3b1ed8d03e0bfc7`, clean clone, `swift build -c release`
- macOS 15, Apple Silicon, Parakeet TDT v3
- `fluidaudiocli transcribe <clip> --streaming`

## Repro 1 — trailing words dropped (clip attached)

14.5s, 16 kHz mono. Regenerate it exactly rather than relying on an attachment:

```bash
curl -s https://api.openai.com/v1/audio/speech \
  -H "Authorization: Bearer $OPENAI_API_KEY" -H "Content-Type: application/json" \
  -d '{"model":"tts-1-hd","voice":"echo","input":"This would go for self correction, this would go for building lists, this would go for places where we would want paragraphs. This would go for where we would make grammatical fixes. This would go for where we would remove filler words."}' \
  -o repro.mp3
afconvert -f WAVE -d LEI16@16000 -c 1 repro.mp3 repro.wav
```

The text repeats one phrase five times:

> This would go for self correction, this would go for building lists, this would go for places where we
> would want paragraphs. This would go for where we would make grammatical fixes. This would go for where
> we would remove filler words.

```
batch (same binary, same clip — reference for what was spoken):
  ...This would go for where we would remove filler words,

--streaming on main @ c7b13a39:
  ...This would go for where we would remove.                 <-- "filler words" dropped

--streaming, same build, ONE line changed:
  ASRConstants.duplicateFrameTolerance = Int.max / 4          (gate effectively off)
  ...This would go for where we would remove filler words.    <-- restored
```

Deterministic: 3/3 runs per arm, byte-stable within each arm. The one-line mutation is the entire diff
between the two arms.

[...]

PR #804, opened 2026-07-17
perf: improve performance

yanxxl:
### Why is this change needed?

Paraformer processing, with the addition of a timestamp-enabled method.
The Paraformer model's performance was improved, boosting the speed on my M4 chip from 48× real-time to 140× real-time.
The SenseVoiceSmall model, despite its claimed high performance, initially ran at only 5× real‑time on my Mac mini M4; later, after optimization, it reached 150× real‑time.
The core code was written by AI, but after my careful testing, the results indeed match the description above.

PR #861, opened 2026-08-19
fix(asr/streaming): re-decode final window from frame 0 — stops #855 trailing-word loss

Alex-Wengg:
Addresses the trailing-word loss from #855 (repro 1).

## Root cause

Repro 1 was reproduced locally from the issue text with macOS TTS. Streaming deterministically dropped `"remove filler words."` while batch transcription retained it.

Tracing showed that the temporal deduplication gate was not the direct cause: the final window's hypothesis reached deduplication nearly empty. The failure occurs when entering the decoder for the short final flush window:

- The window contains left context plus the remaining audio (about 44 encoder frames in the repro).
- When `timeJump == 0`, the normal continuation path skips the standard two-second overlap and starts at frame 25.
- Entering that short window mid-stream with the carried LSTM state emits boundary punctuation and then blanks through the remaining frames, dropping clear trailing speech.
- Decoding the same window from frame 0 produces the missing lexical tokens.

## Exact WAV reproduction

No API key or external TTS service is required. This command uses macOS TTS and writes 16-bit little-endian PCM at 16 kHz; the `.wav` extension selects the WAVE container:

```bash
say -o /tmp/repro855.wav --data-format=LEI16@16000 "This would go for self correction, this would go for building lists, this would go for places where we would want paragraphs. This would go for where we would make grammatical fixes. This would go for where we would remove filler words."
```

The generated file was verified with `afinfo /tmp/repro855.wav`:

```text
File type ID:   WAVE
Data format:    1 ch, 16000 Hz, Int16
Duration:       12.485250 s
```

For the verification run documented below, the artifact was 403,624 bytes with SHA-256:

```text
b0ec1e67653b6432bce96eaea0362676cd6cc5cf4f09210e70e680745ec73fb6
```

The command does not specify `-v`, so it uses the configured default macOS voice. The container and PCM properties are fixed, but duration and bytes can vary with the installed/default voice and OS version.

Using the same WAV for every run:

```bash
swift run -c release fluidaudiocli transcribe /tmp/repro855.wav --streaming --model-version v3
swift run -c release fluidaudiocli transcribe /tmp/repro855.wav --model-version v3
```

Observed transcript tails:

| Revision/mode | Result |
|---|---|
| Unfixed PR base `df1417ce`, streaming | `…this would go for where we would.` — dropped `remove filler words.` |
| Fixed head `2a522d9b`, streaming | `…this would go for where we would. remove filler words.` — lexical tail restored |
| Fixed head `2a522d9b`, batch reference | `…this would go for where we would remove filler words.` |

The extra sentence-final period in the fixed streaming result is the known cosmetic boundary behavior described below.

## Fix

For the final streaming window with accumulated tokens and timestamps, `lastChunkRedecodePlan` now:

1. Sets `initialTimeIndexOverride: 0` so decoding starts at the beginning of the window.
2. Derives a window-local emission cutoff from the latest accumulated token timestamp.
3. Backs the cutoff up by five frames to allow for cross-window emission jitter.

The decoder still processes the suppressed overlap and updates its LSTM state, but it does not append tokens for audio already covered by previous windows. Temporally gated deduplication therefore handles only the small jitter margin plus genuinely new trailing content, rather than the entire re-decoded overlap. This also prevents token-dense overlaps from exceeding the deduplicator's bounded search window.

The override and emission cutoff are forwarded through all supported decoder routes, including Parakeet v2 and TDT-CTC 110m as well as v3 and tdtJa.

Batch `ChunkProcessor`, interior streaming windows, and single-window streams without accumulated history keep their existing navigation.

## Verification

Verified on an M5 Pro using a release build:

| Clip | Before | After |
|---|---|---|
| #855 macOS TTS WAV | Trailing `"remove filler words"` dropped | Restored, 3/3 stable |
| 26-second multi-window clip | — | All content present |
| FLEURS `en_us_0000` | — | Batch and streaming byte-identical |
| Two voice variants | One clean; one split-word artifact | Clean / content-complete |

The manual audio verification used Parakeet TDT v3. The v2/110m forwarding is covered by the shared decoder route in this change.

## Tests

- Captured #855 trace: strips the re-decoded overlap, including its partial leading token, while retaining the trailing words.
- Timestamp inverse: keeps an identical token run when the occurrences are about ten seconds apart.
- Final-window decoder-entry plan: verifies the frame-0 override and emission-cutoff arithmetic.
- Cutoff clamping and inactive cases: verifies legacy navigation for interior windows, callers without timestamps, and first/single windows.

## Known remaining behavior

- Repro 2 from #855 (a phantom phrase at an interior seam) is not directly addressed because the reporter's private clip is required for verification.
- A truncated window can still emit sentence-final punctuation at its boundary, producing cosmetic output such as `"…we would. remove filler words."`. This pre-existing behavior is intentionally not trimmed heuristically because timestamp-based trimming can remove real words.

### DIFF
--- a/Sources/FluidAudio/ASR/Parakeet/SlidingWindow/TDT/AsrManager+Transcription.swift
+++ b/Sources/FluidAudio/ASR/Parakeet/SlidingWindow/TDT/AsrManager+Transcription.swift
@@ -52,6 +52,34 @@ extension AsrManager {
         return result
     }
 
+    /// Cross-window emission jitter allowance for the final-window re-decode: a
+    /// re-decoded token can land a few frames from its original emission, so the
+    /// suppression cutoff backs off this much and dedup strips what remains.
+    internal static let redecodeEmissionJitterFrames = 5
+
+    /// Decoder-entry plan for the final streaming window (issue #855).
+    ///
+    /// Returns `initialTimeIndexOverride: 0` so the decoder re-decodes the window
+    /// from frame 0 with its carried state (a mid-window entry into a short flush
+    /// window can blank out the trailing speech), plus an emission cutoff in
+    /// window-local frames: tokens for audio the previous windows already emitted
+    /// are suppressed at the source, leaving dedup only the jitter margin.
+    /// Non-final windows and callers without accumulated timestamps get `(nil, nil)`
+    /// — the legacy navigation.
+    nonisolated internal static func lastChunkRedecodePlan(
+        isLastChunk: Bool,
+        previousTokens: [Int],
+        previousTokenTimestamps: [Int]?,
+        globalFrameOffset: Int
+    ) -> (initialTimeIndexOverride: Int?, emitTokensAfterFrame: Int?) {
+        guard isLastChunk, let previousTimestamps = previousTokenTimestamps, !previousTokens.isEmpty else {
+            return (nil, nil)
+        }
+        let lastEmittedGlobalFrame = previousTimestamps.max() ?? 0
+        let cutoff = max(0, lastEmittedGlobalFrame - globalFrameOffset - redecodeEmissionJitterFrames)
+        return (0, cutoff)
+    }
+
     /// Chunk transcription that preserves decoder state between calls.
     /// Used by SlidingWindowAsrManager for overlapping-window processing with token deduplication.
     func transcribeChunk(
@@ -70,9 +98,16 @@ extension AsrManager {
         // Jumping mid-window into a short flush window can blank out the trailing
         // speech entirely (issue #855: the joint emits a boundary punctuation, then
         // blanks to the end, dropping the final words). Re-decoding the overlap with
-        // the carried state is robust, and the temporally-gated dedup below strips
-        // the re-decoded tokens (they land within a few frames of the originals).
-        let redecodeFromWindowStart = isLastChunk && previousTokenTimestamps != nil && !previousTokens.isEmpty
+        // the carried state is robust; emissions for audio the previous windows
+        // already covered are suppressed at the source (the decoder still updates
+        // its LSTM state through them), so dedup only sees the few-frame jitter
+        // margin — a token-dense overlap cannot outgrow dedup's bounded search.
+        let redecodePlan = Self.lastChunkRedecodePlan(
+            isLastChunk: isLastChunk,
+            previousTokens: previousTokens,
+            previousTokenTimestamps: previousTokenTimestamps,
+            globalFrameOffset: globalFrameOffset
+        )
         let (hypothesis, encLen) = try await executeMLInferenceWithTimings(
             padded,
             originalLength: frameAlignedLength,
@@ -81,7 +116,8 @@ extension AsrManager {
             contextFrameAdjustment: 0,  // Non-streaming chunks don't use adaptive context
             isLastChunk: isLastChunk,
             language: language,
-            initialTimeIndexOverride: redecodeFromWindowStart ? 0 : nil
+            emitTokensAfterGlobalFrame: redecodePlan.emitTokensAfterFrame,
+            initialTimeIndexOverride: redecodePlan.initialTimeIndexOverride
         )
 
         // Apply token deduplication if previous tokens are provided
--- a/Sources/FluidAudio/ASR/Parakeet/SlidingWindow/TDT/AsrManager+Transcription.swift
+++ b/Sources/FluidAudio/ASR/Parakeet/SlidingWindow/TDT/AsrManager+Transcription.swift
@@ -66,6 +66,13 @@ extension AsrManager {
         let (alignedSamples, frameAlignedLength) = frameAlignedAudio(
             chunkSamples, allowAlignment: previousTokens.isEmpty)
         let padded = padAudioIfNeeded(alignedSamples, targetLength: ASRConstants.maxModelSamples)
+        // Last streaming window: decode from frame 0 instead of skipping the overlap.
+        // Jumping mid-window into a short flush window can blank out the trailing
+        // speech entirely (issue #855: the joint emits a boundary punctuation, then
+        // blanks to the end, dropping the final words). Re-decoding the overlap with
+        // the carried state is robust, and the temporally-gated dedup below strips
+        // the re-decoded tokens (they land within a few frames of the originals).
+        let redecodeFromWindowStart = isLastChunk && previousTokenTimestamps != nil && !previousTokens.isEmpty
         let (hypothesis, encLen) = try await executeMLInferenceWithTimings(
             padded,
             originalLength: frameAlignedLength,
@@ -73,7 +80,8 @@ extension AsrManager {
             decoderState: &decoderState,
             contextFrameAdjustment: 0,  // Non-streaming chunks don't use adaptive context
             isLastChunk: isLastChunk,
-            language: language
+            language: language,
+            initialTimeIndexOverride: redecodeFromWindowStart ? 0 : nil
         )
 
         // Apply token deduplication if previous tokens are provided
--- a/Sources/FluidAudio/ASR/Parakeet/SlidingWindow/TDT/AsrManager.swift
+++ b/Sources/FluidAudio/ASR/Parakeet/SlidingWindow/TDT/AsrManager.swift
@@ -300,7 +300,9 @@ public actor AsrManager {
                 decoderState: &decoderState,
                 contextFrameAdjustment: contextFrameAdjustment,
                 isLastChunk: isLastChunk,
-                globalFrameOffset: globalFrameOffset
+                globalFrameOffset: globalFrameOffset,
+                emitTokensAfterGlobalFrame: emitTokensAfterGlobalFrame,
+                initialTimeIndexOverride: initialTimeIndexOverride
             )
         case .v3:
             // Pass `vocabulary` unconditionally. `TdtDecoderV3.tokenLanguageFilter`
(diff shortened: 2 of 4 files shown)


saurabhav88:
Ran this branch against our corpus. Repro 2 is **not** fixed, and on real dictation the change is a net regression, so please do not read this as a green light.

Method: `fluidaudiocli transcribe <clip> --model-version v3 [--streaming]`, three arms each built from its own clean clone: `main` @ `df1417ce`, this PR @ `b93d34ca`, and our fork @ `bf9fe27f` for context. Corpus is 500 archived real dictations, frozen to a membership manifest with a per-file sha256 that is re-verified before each decode. Batch decode is the reference for what was spoken. Machine: M5 Max, macOS 26.4, release builds.

## Repro 2: unchanged, 3/3

The 17.2 s dictation from the issue, counting occurrences of the repeated phrase:

| build | count | |
|---|---|---|
| batch (reference) | 5 | correct |
| our fork @ `bf9fe27f` | 5 | correct |
| `main` @ `df1417ce` | **6** | phantom `This would go for,` at the seam |
| this PR @ `b93d34ca` | **6** | unchanged, deterministic 3/3 |

Byte-identical across three runs. The phantom sits in the same place.

## Repro 1: depends on which clip

You regenerated the issue text with `say`. I ran both that and the OpenAI `tts-1-hd` / `echo` clip the issue documents.

| clip | `main` | this PR |
|---|---|---|
| `say`, 12.5 s | drops `remove filler words` | **restored** (with the cosmetic stray `.` you noted) |
| OpenAI `tts-1-hd` echo, 14.7 s (the issue's own clip) | drops `remove filler words` | **still dropped** |

So the fix is real on one clip and does not reach the other. Both are the same sentence; only the voice differs.

## 500 real dictations: 49 changed, and most changed for the worse

493 clips decoded in all three arms (a handful rotated out of the live archive mid-run and are reported as gaps, not skipped). This arm was run twice end to end: 489 comparable clips, zero differences between runs, so nothing below is run-to-run noise.

`main` vs this PR: **49 clips differ**. Scoring both against the batch reference of the same clip, on a case- and punctuation-folded token stream:

| verdict | clips |
|---|---|
| PR closer to what was spoken | 11 |
| **PR further from what was spoken** | **30** |
| tie | 3 |
| punctuation only | 5 |

Total word errors across those 49 clips: **`main` 211 → PR 317**, against 2,670 reference words.

The wins are the bug you targeted, and they are genuine. `main` ends `That is why we are using.` where the PR recovers `That is why we are using our Azure credits and we are using LLMs to provide us that feedback.`

The losses share a shape: the final window gains leading context and loses its tail.

```
batch: ...the netted weighted score is usually lower um and we haven't reached the level of
       confidence of like high quality polish in these categories.
main : ...the netted weighted score is usually lower uh um and we haven't reached the level of
       confidence of like high quality polish in these categories.
PR   : ...the netted weighted score is usually lower uh um and we haven't reached the levelity.
```

```
batch: ...we could potentially have multiple audio files being stored if I just keep canceling
       over and over and over again.
main : ...we could potentially have multiple audio files being st stored if I just keep canceling
       over and over and over again.
PR   : ...we could potentially have multiple audio files being still.
```

Same in `Can we pull it up live and show them the configuration in Qualtrics?` → `Can wear it?`, and `select specific Apple intelligence models` → `selectum`. These are 5 to 14 word losses on clips `main` transcribes correctly, so the re-decode is trading a rare tail loss for a more common one.

## One more datapoint on where to look

The one-line isolation from the issue no longer isolates anything on current `main`. Setting `ASRConstants.duplicateFrameTolerance = Int.max / 4` leaves all three clips byte-identical to stock `main`, including repro 2, which it used to restore to 5 at `c7b13a39`. That mutation is live, not inert: setting the same constant to `0` visibly changes output (repro 2 becomes `This would go forxes, this would go for...`).

Three ASR commits landed between `c7b13a39` and `df1417ce`: #709, #804, and #857, whose summary mentions a timestamp tail. We have not isolated which, and we are not claiming one. But if the gate was the whole story in the original report and is not the whole story now, repro 2's cause may have moved, which would explain why a decoder-entry fix does not touch it.

Happy to run any patch through the same corpus. It is a one-command turnaround at this end.

saurabhav88:
Follow-up to my previous comment. We put that analysis through an independent adversarial review before doing anything further with it, and the review was right about two overclaims on our side. Corrections first, then the extra evidence it produced, which is more useful to you than the original numbers were.

## Two corrections to what I wrote

**1. I called batch decode "the reference for what was spoken". It is not.** Batch is a machine transcript from the same model, so it is a proxy, not ground truth. Nobody has listened to these 49 clips and transcribed them by hand. Everywhere I wrote "closer to what was spoken", please read "closer to the batch proxy". The direction of the result is unaffected, but "net accuracy regression" is stronger than what we measured, and I should have written "a substantial regression signal against a batch proxy, worth review before merge".

**2. I suggested repro 2's cause "may have moved" since `c7b13a39`.** That was speculation dressed as an inference. What we actually established is narrower: the old one-line intervention no longer changes output on current `main`. That is a null result, and a null result does not identify a cause. Other explanations fit it equally well, including that our original diagnosis was incomplete. A bisect of `c7b13a39..df1417ce` on repro 2, run with both stock and max-tolerance builds, would settle it. Please do not spend time on #709, #804, or #857 on our say-so.

Everything factual in the first comment reproduced exactly on recount, including the 49 changed clips and the 11 closer / 30 further / 3 tie / 5 punctuation-only split.

## The oracle is not contaminated by our fork

The obvious objection to the first comment was that our batch reference came from our patched build. We tested it rather than argued about it: the corpus was decoded in batch again on stock `main` @ `df1417ce`.

- 484 clips decoded in batch on both arms. The two arms disagree on **1**, and that clip is not among the 49.
- On the 49 changed clips, main batch was available for 48, and fork batch and main batch are **byte-identical on all 48**.
- Scoring against main batch instead of fork batch gives 11 closer, 29 further, 3 tie, 5 punctuation-only. The one clip that moves is the one that rotated out of our live archive between runs.

So the fork carries did not shape the verdict.

## Punctuation folding was hiding some of it, not creating it

Our normalisation folded away case and punctuation. That is a real limitation, so we ran the same adjudication three ways:

| tokenisation | closer | further | tie |
|---|---|---|---|
| case and punctuation folded (what we published) | 11 | 30 | 3 |
| lowercased, punctuation kept | 11 | 32 | 5 |
| raw whitespace tokens | 12 | 32 | 5 |

The folding softens the negative result rather than manufacturing it.

## Where the changes actually land, which is the part likely to help you

Of the 30 clips that score worse against the proxy:

- **28 first diverge from `main` within the last 20 words.** The effect is concentrated in the final window, which is what the change targets.
- **17 produce a shorter transcript than `main`.**
- **Only 2 are an exact prefix of `main`.** So this is not simple truncation. Several of the rest add repeated or extra leading context while keeping the tail.

I should also correct a mechanism claim I made. I wrote that the final window "gains leading context and loses its tail". The change does not remove tail audio, and the same samples are still available to the decoder. Starting at frame 0 with carried state can alter the decode trajectory, re-emit context, and interact differently with dedup, and a per-chunk token budget consumed before the tail is reached would produce what we see. That is consistent with the change, but the trade is an empirical side effect, not the mechanism the patch implements. Treat the shape as a symptom to trace, not as a diagnosis from us.

## One suggestion on the tests

The regression tests added here exercise dedup over prepared token and timestamp arrays. That is the right unit for the dedup gate, and it cannot see this failure class, because the failure appears only when a real final-window decode runs end to end. A test that feeds real audio windows through `finish()` and asserts on the resulting text would have caught the 49-clip effect. We would have caught it earlier ourselves if we had one.

## Standing offer

We will run any patch through the same corpus and report back the same way. Turnaround is a single command here, about twenty minutes for a full three-arm sweep, and we can report per-clip diffs rather than just totals.

Thanks for the quick turnaround on this one, and sorry for the extra work the corrections above create.

saurabhav88:
Repro 2 is now attachable. We cleared the one clip for release, so you no longer have to work from our description of it.

**https://github.com/saurabhav88/FluidAudio/releases/tag/issue-855-repro2**

`issue-855-repro2-17.2s.wav`, 17.197 s, 16 kHz mono 16-bit PCM, sha256 `80b9e000d45b0a6f8153423090f81eff88ab88d928b509f869875114d94abbe3`. It is the exact buffer our pipeline handed to the decoder, so there is no resampling step between you and the failure.

On why this one and not the corpus: the spoken content is a scripted test phrase recorded for this purpose, with nothing personal in it. That is what made it clearable. The other 499 are real user dictation and stay on the machine.

Verified end to end against the published copy rather than our local one, by downloading the release asset back and decoding that file:

| build | mode | occurrences of `would go for` |
|---|---|---|
| `main` @ `df1417ce` | batch | 5 |
| `main` @ `df1417ce` | streaming | **6** |
| this PR @ `b93d34ca` | streaming | **6** |
| our fork @ `bf9fe27f` | streaming | 5 |

```
fluidaudiocli transcribe issue-855-repro2-17.2s.wav --model-version v3
fluidaudiocli transcribe issue-855-repro2-17.2s.wav --model-version v3 --streaming
```

The sixth occurrence is `This would go for,` inserted at a chunk seam, between `grammatical fixes.` and `This would go for where we would remove filler words.` Batch decode of the same file, same binary, gives five.

Two things this clip is good for beyond reproducing the bug. It is a real recording rather than synthesised speech, which matters here because the trailing-word behaviour in your #861 verification differed between a `say` clip and an OpenAI TTS clip of the identical sentence. And it is small enough to sit in a repository as a fixture, so a real-audio streaming test could assert on it directly, which is the gap I mentioned in my previous comment.

Happy to record and clear further clips against a specific shape if that would help. Describe the utterance pattern you want and we will produce it.

PR #884, opened 2026-08-31
fix(tts/kokoro-ane): stem possessives instead of G2P-ing the whole token

akmittal006:
## Problem

Kokoro ANE sends ordinary possessives absent from Misaki's lexicon to neural G2P as whole tokens, which can produce incorrect pronunciations. Derive known stems with the appropriate /s/, /ᵻz/, or /z/ clitic while preserving explicit dictionary pronunciations.

## Resolution order

- Explicit entries for the full inflected token retain precedence.
- For hyphenated possessives, look up the whole bare stem directly in the custom and bundled lexicons. This preserves whole-compound stress and reduced vowels in `C-section's` and `mother-in-law's`. The probe cannot recurse, split, or invoke G2P.
- On a whole-stem miss, split the compound and resolve each component normally. This preserves explicit `use's` and `produce's` noun readings in `land-use's` and `fresh-produce's`.
- Ordinary possessives resolve known stems without neural fallback, retaining custom overrides and initialism letter names. Unknown stems keep the original inflected token for fallback.

Apostrophe variants are folded before tokenization. The change is confined to possessive `'s`; it does not add general plural stemming.

## Validation

- 70 related Kokoro ANE and StyleTTS2 phonemizer tests pass, including 50 Kokoro English tests.
- New whole-compound regressions fail on the preceding PR head and pass after the follow-up fix.
- Coverage includes whole compound pronunciations, heteronym components, explicit inflected entries, custom/case-sensitive precedence, apostrophe variants, initialisms, and OOV fallback arguments.
- A separate local audit against the real cached US lexicon passed for 3,726 tokenizable compound stems and 372 explicit possessives, without G2P fallback. The local-cache audit is not part of CI.
- Strict formatting lint and `git diff --check` pass.

Validation covers phoneme resolution; no new audio listening or end-to-end synthesis evaluation is claimed.

### DIFF
--- a/Sources/FluidAudio/TTS/KokoroAne/G2P/English/KokoroAneEnglishPhonemizer.swift
+++ b/Sources/FluidAudio/TTS/KokoroAne/G2P/English/KokoroAneEnglishPhonemizer.swift
@@ -16,14 +16,13 @@ import Foundation
 ///      citation form (`tˈO`) that over-stresses them (issue #691)
 ///   6. strict ASCII all-caps initialisms (`FBI`, `ATP`) spelled as
 ///      letter names after a full lexicon miss (issue #710)
-///   7. hyphenated-compound split for a token that missed as a whole
-///      (`land-use's` → `land` + `use's`), resolving each part through
-///      this same chain (issue #775)
-///   8. `-'s` stem + clitic for possessives whose stem is a known word
-///      (`today's` → `today` + /z/), mirroring Misaki's `stem_s`. It runs
-///      *after* the hyphen split so a compound's final part can still hit
-///      a glued lexicon entry of its own
-///   9. BART G2P CoreML fallback for OOV words (injected by the caller)
+///   7. whole-compound possessive stem lookup, using lexicons only
+///      (`C-section's` → lexicon `C-section` + /z/)
+///   8. hyphenated-compound split after a whole-stem miss
+///      (`land-use's` → `land` + lexicon `use's`) (issue #775)
+///   9. `-'s` stem + clitic for other known stems (`today's` → `today` + /z/),
+///      including letter-name initialisms (`FBI's`)
+///   10. BART G2P CoreML fallback for OOV words (injected by the caller)
 ///
 /// Punctuation supported by the chain's `vocab.json` (`, . ! ? ; …` etc.)
 /// is preserved and attached to the preceding word — Kokoro treats those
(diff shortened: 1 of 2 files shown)


Alex-Wengg:
Calling `resolvePossessive` before hyphen splitting introduces a lexicon-precedence regression when the full compound misses the lexicon.

For example, `land-use's` previously resolved `land` + `use's`, preserving the noun pronunciation. It now resolves `land` + `use` and appends the clitic, picking the verb pronunciation of "use." Likewise, `fresh-produce's` now uses the verb pronunciation of "produce."

I reproduced both against the parent commit using the real cached lexicon. Could we preserve explicit possessive entries for compound components before deriving their pronunciation, and add regression tests for these cases?

akmittal006:
Good catch, and thank you for reproducing it against the real cache. The lexicon does carry glued `'s` keys (347 in the US cache), nearly all noun/verb heteronyms like `use's` and `produce's`, and my ordering let the stem + clitic derivation outrank them for a compound's final part.

The follow-up commit splits hyphenated compounds before the possessive rule, so a part with its own glued entry wins (`land-use's` → `lˈænd jˈusᵻz`, `fresh-produce's` → `fɹˈɛʃ pɹˈOdˌusᵻz`) while a part with none still stems (`mother-in-law's` unchanged). Both of your examples are now regression tests that fail on the previous head. I've also corrected the PR description, which wrongly said no glued possessive keys exist.

Alex-Wengg:
Thanks, the `land-use's` ordering is right, but splitting first breaks the opposite case, which is far more common. The real cache has 3,731 hyphenated whole-word stems (`C-section`, `X-ray`, `T-shirt`, `well-being`, `mother-in-law`) and exactly one glued hyphenated possessive (`re-count's`). So `C-section's` now resolves as `C` + `section's` instead of the compound stem `sˈisˌɛkʃən` + `z`. Your `mother-in-law's` example hits it too: the whole stem is in the lexicon with reduced vowels, and the split path gives three full-vowel words. Misaki's `stem_s` tries the whole stem first.

Add one step between the whole-token miss and the hyphen split: if the token ends in `'s`, look the stem up as a whole against the lexicons only (no split, no G2P fallback) and append the clitic on a hit. `land-use's` still misses there and falls through to the split, where `use's` wins. Regression tests for `C-section's` and `mother-in-law's` would lock it in. I've approved the CI runs on your head.

akmittal006:
Addressed in 85d1f8ab. After a whole inflected-token miss, hyphenated possessives now probe the whole stem directly in the custom and bundled lexicons, preserving their existing lookup precedence. This probe cannot split, recurse, or call G2P. Only a whole-stem miss proceeds to component resolution, so `land-use's` and `fresh-produce's` retain their explicit noun-possessive entries.

Added regressions for `C-section's`, `mother-in-law's`, `X-ray's`, `T-shirt's`, and `well-being's`, including apostrophe variants. Also covered case-sensitive/custom precedence, the explicit `re-count's` entry, OOV fallback arguments, and `AI's`/`US's`/`FBI's`. The early probe is limited to hyphenated stems so ordinary initialisms keep their letter-name overrides.

Validation: the new whole-compound regressions failed on the previous head and pass with this change; all 70 related Kokoro ANE and StyleTTS2 phonemizer tests pass. A separate local sweep against the real cached US lexicon passed for 3,726 tokenizable compound stems and 372 explicit possessives, with no G2P fallback. Formatting and whitespace checks also pass. This validates phoneme resolution; it is not an audio listening or end-to-end synthesis claim.

PR #890, opened 2026-09-08
Build the library for tvOS: make NemoTextProcessing platform-conditional

cakesmachina:
The prebuilt `NemoTextProcessing.xcframework` ships ios, ios-simulator and macos slices only, so building the `FluidAudio` library for tvOS fails with "no library for this platform was found". The framework is imported by exactly two files and used at two TTS call sites; the Parakeet ASR pipeline never touches it.

This PR:
- makes the binary dependency `.when(platforms: [.macOS, .iOS])` in `Package.swift`;
- wraps `ITN/TextNormalizer.swift` and `TTS/Shared/NemoTextNormalizer.swift` in `#if canImport(CNemoTextProcessing)`;
- lets `EnglishTextNormalizer.normalizeForFrontend` fall back to its baseline and the Mandarin Kokoro path pass text through when the FST engine is absent.

Witnessed with Xcode 27 beta: `swift build --product FluidAudio --triple arm64-apple-tvos26.0-simulator` and `--triple arm64-apple-tvos26.0` both complete. macOS and iOS builds are unchanged. `platforms:` in `Package.swift` is deliberately left alone; adding `.tvOS` there is your call.

Use case: bundling `parakeet-tdt-0.6b-v3-coreml` inside an Apple TV app for offline read-along transcripts.

https://claude.ai/code/session_01HZbYTEH3aemaaG2otBWesj

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -25,7 +25,7 @@ let package = Package(
             dependencies: [
                 "FastClusterWrapper",
                 "MachTaskSelfWrapper",
-                "NemoTextProcessing",
+                .target(name: "NemoTextProcessing", condition: .when(platforms: [.macOS, .iOS])),
             ],
             path: "Sources/FluidAudio",
             exclude: ["ASR/Parakeet/Unified/benchmark.md"],
--- a/Sources/FluidAudio/ITN/TextNormalizer.swift
+++ b/Sources/FluidAudio/ITN/TextNormalizer.swift
@@ -1,3 +1,4 @@
+#if canImport(CNemoTextProcessing)
 import CNemoTextProcessing
 import Foundation
 import NaturalLanguage
@@ -254,3 +255,5 @@ public final class TextNormalizer: Sendable {
         return out
     }
 }
+
+#endif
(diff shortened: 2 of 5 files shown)


Alex-Wengg:
Thanks. Conditional binary dependency is the right call (confirmed v0.3.0 ships only macOS/iOS slices). Two fixes before merge:

**1. Don't remove `TextNormalizer` / `NemoTextNormalizer` on tvOS.** Wrapping the whole files drops two public types, so anything calling `TextNormalizer.shared` won't compile on tvOS. It also breaks the documented contract (`PostProcessing.md`): API always present, `isNativeAvailable` reports linkage, `normalize()` passes through otherwise. Guard only the import and the FFI call bodies, return input with `isNativeAvailable == false`. Then the `#if` blocks in `KokoroAneManager` and `EnglishTextNormalizer` can go, since `NemoTextNormalizer.normalize` already returns `text` unchanged when the engine declines.

**2. Mandarin numeric-only input.** `MandarinG2P` already runs `MandarinNumberNormalizer`, but `looksLikeHanzi` gates before it. Without the FST, `$5.50` / `99%` / `2025-05-03` have no Hanzi, hit the bopomofo passthrough, and digits become tone marks. When the FST is a no-op and there's no Hanzi, run `MandarinNumberNormalizer.normalize` before the gate.

Also:
- Add a tvOS build to CI (`swift build --product FluidAudio --triple arm64-apple-tvos-simulator`); the x86_64 job from #876 is a template. Nothing currently compiles that triple.
- Add `.tvOS(.v17)` to `platforms:` so the deployment target is explicit.

PR #891, opened 2026-09-08
fix(offline-diarizer): hold speaker caps against both cluster censuses

ertra:
> **Updated after review.** The first version filtered the centroid census, which
> changed the unconstrained default path — thanks @Alex-Wengg for catching that. This
> version takes the route you described: centroids stay on the pi census, and the gate
> in `refineWithConstraints` checks both counts. Rebased onto current `main`.

## Problem

`OfflineDiarizerManager` can return more speakers than `clustering.numSpeakers` or
`clustering.maxSpeakers` allows. The count correction concludes no adjustment is needed, and a
later stage reintroduces a cluster that conclusion had already excluded.

Deterministic counterexample, no audio or models needed:

- `gamma = [[0.6, 0.4], [0.6, 0.4]]`, `pi = [0.6, 0.4]`, `numClusters = 2`
- two local speakers sharing one segmentation chunk
- requested `numSpeakers = 1` (or `maxSpeakers = 1`)

Expected one speaker; the assignment comes back `[0, 1]` — two.

## Why

Three stages currently use three different definitions of a surviving cluster:

1. The count check reads **argmax winners** — `output.assignedClusterCount`
   ([VBxClustering.swift#L703](https://github.com/FluidInference/FluidAudio/blob/b980fc2e/Sources/FluidAudio/Diarizer/Offline/Clustering/VBxClustering.swift#L703)).
   Column 1 wins no argmax, so the detected count is 1, `needsAdjustment` is false, and
   `refineWithConstraints` returns without re-clustering.
2. Centroid construction keeps **every cluster with `pi > 1e-7`**
   ([OfflineDiarizerManager.swift#L701](https://github.com/FluidInference/FluidAudio/blob/b980fc2e/Sources/FluidAudio/Diarizer/Offline/Core/OfflineDiarizerManager.swift#L701)),
   so column 1 gets a centroid anyway.
3. Constrained assignment is enabled whenever the count was not adjusted and more than one
   centroid exists
   ([OfflineDiarizerManager.swift#L422](https://github.com/FluidInference/FluidAudio/blob/b980fc2e/Sources/FluidAudio/Diarizer/Offline/Core/OfflineDiarizerManager.swift#L422)).
   Its pyannote-parity rule forces two speakers sharing a chunk onto distinct clusters, so the
   second speaker lands on the revived column.

Setting `clustering.constrainedAssignment = false` does not close it — plain cosine assignment
over the same retained centroids can select the non-winning column too.

## This change

Stage 2 is deliberate pyannote parity (`VBxClustering.__call__` keeps every `sp > 1e-7`
component and assigns over all of them), so the fix belongs in stage 1: **re-cluster when
either census falls outside the bounds.**

```swift
let assignedCount = output.assignedClusterCount
let activeCount = output.activeClusterCount
let detectedCount =
    constraints.needsAdjustment(detectedCount: assignedCount) ? assignedCount : activeCount
guard constraints.needsAdjustment(detectedCount: detectedCount) else { return output }
```

The argmax census leads when both are out of bounds, so an under-count still re-clusters to the
target #802 chose and reports the same detected count; the pi census only decides cases the
argmax census called satisfied. It also catches a pi census above `maxSpeakers`.

Unconstrained runs return before the gate, so nothing about the default path moves — centroid
construction is untouched.

## Two extractions, to make it testable

The gate sits below a PLDA-model-dependent refinement, so there was no model-free path from a
`VBxOutput` to actual assignments — which is why the first version's tests could not see the
regression they introduced. Both moves are pure, no behaviour change:

- `VBxClustering.applyConstraints(to:trainingEmbeddings:constraints:)` — the constraint gate,
  static, split out of `refineWithConstraints`.
- `OfflineDiarizerManager.clusterAssignments(...)` — centroid construction plus the assignment
  rule, split out of `cluster(_:)`.

They exist only to make the tests possible. Happy to reshape them if you would rather the seams
were drawn differently — different names, a different split, an injectable instead of a static.

## Tests

New `OfflineSpeakerCapTests` asserts the speaker count a caller actually observes,
`Set(assignments).count`, by running the real assignment:

- `numSpeakers: 1` and `maxSpeakers: 1` on the toy case each yield 1 speaker. Both fail on
  `main` with `("2") is not equal to ("1")`.
- **`testUnconstrainedRunKeepsBothMixtureComponents`** — the same input with no request must
  still yield 2. Passes before and after; it is the regression the first version would have
  caused.
- `#802` guard: an argmax under-count still re-clusters and reports that census as detected.
- `#801` guard: when both censuses fit, K-Means does not run.

`VBxConstraintTests` 10/10. Full diarizer/clustering selection: 236 tests, 7 failures — all 7 in
`SpeakerEnrollmentTests` (`inferenceFailed("Missing chunk_pre_encoder_embs_out")`), identical on
an unmodified `main` checkout here, so pre-existing and unrelated. `swift build --arch x86_64`
clean.

## Relationship to #802

This is a side effect of #802, which is otherwise a clear improvement: the count check became
more accurate, but the downstream stages were not moved to the same definition.

One detail worth stating precisely, because it is easy to misread. Before #802 the check read
`output.numClusters`, which at that point was `max(1, Set(initialClusters).count)` — the **AHC
warm-start count**, not the surviving mixture-weight census. With 2 initial clusters against a
request of 1, `needsAdjustment` was true and K-Means re-clustered to 1, so the constraint held.
#802 rightly stopped gating on the warm-start count (it re-clustered even when VBx already
agreed, #801). The gap is only that the other definitions remained downstream.

## Context

Found while qualifying an upgrade from v0.15.5 for an app with a user-facing "at most N other
speakers" control, where exceeding the limit is visible to the user. I have not measured how often
this occurs on real audio; the report is source analysis plus the deterministic case above.

### DIFF
--- a/Sources/FluidAudio/Diarizer/Offline/Clustering/VBxClustering.swift
+++ b/Sources/FluidAudio/Diarizer/Offline/Clustering/VBxClustering.swift
@@ -27,6 +27,7 @@ struct VBxClustering {
     private let config: OfflineDiarizerConfig
     private let pldaTransform: PLDATransform
     private let logger = AppLogger(category: "OfflineVBx")
+    private static let constraintLogger = AppLogger(category: "OfflineVBx")
     private let signposter = OSSignposter(
         subsystem: "com.fluidaudio.diarization",
         category: .pointsOfInterest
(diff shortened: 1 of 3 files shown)


Alex-Wengg:
Thanks, the counterexample and the three-census diagnosis are right. The fix goes too far though: `centroidClusterIndices` also drops non-winning components in the unconstrained path. pyannote (`VBxClustering.__call__`) keeps every `sp > 1e-7` component as a centroid and runs constrained assignment over all of them, so an overlap speaker can land on one that won no argmax. On your toy input with no constraints, pyannote gives 2 speakers; this PR gives 1. That's a default-config parity change the tests can't see since they don't run assignments.

Take the other route instead: keep centroids on the pi census, and in `refineWithConstraints` trigger K-Means when either `assignedClusterCount` or `activeClusterCount` is out of bounds. Fixes your over-count, keeps the #802 under-count fix, catches a pi census above `maxSpeakers` too, and leaves unconstrained runs untouched. A pipeline-level test on the toy case would lock it in.

ertra:
Reworked in 9b2d559e as suggested: centroids keep the pi census, and the constraint gate re-clusters when either count falls outside the bounds. When both do, the argmax census leads, so an under-count still re-clusters to the target #802 chose — say the word if you would rather the pi census won that case.

Added assignment-level tests: one speaker with either cap, two with no constraints. The unconstrained one is the parity case your review flagged, so it is pinned by a test now rather than by argument. The two internal extractions (`applyConstraints`, `clusterAssignments`) exist so those tests can run the real assignment without loading models.

Alex-Wengg (approved):
Reviewed the rework at 9b2d559e. The dual-census gate, target clamping, assignment-level coverage, and unconstrained-path preservation address the prior findings. All required checks pass.

PR #892, opened 2026-09-08
feat(package): make NemoTextProcessing an opt-out trait for ASR-only consumers (#880, #888)

### LINKED ISSUE #880: 0.15.5+ links NemoTextProcessing (a prebuilt Rust staticlib) into every FluidAudio consumer; apps that ship their own Rust core cannot link
Since 0.15.5 the `FluidAudio` target depends on `NemoTextProcessing`, a prebuilt Rust static library used by the TTS path (Kokoro). Apps that already link a Rust static library of their own (we ship a UniFFI core, `libflow_ffi.a`) can no longer link against FluidAudio 0.15.5+:

- `duplicate symbol '_rust_eh_personality'` and 143 other Rust std symbols defined in both archives;
- after localizing our own Rust symbols behind the UniFFI entry points, the app still fails with `ld: Too many personality routines for compact unwind to encode` (two `_rust_eh_personality`, `___gxx_personality_v0`, `___objc_personality_v0`: the compact unwind format allows three), and `-no_compact_unwind` breaks Rust panic propagation (`failed to initiate panic, error 5`).

We do not use TTS. Would you consider making `NemoTextProcessing` an optional product (a separate SwiftPM product or target, or a package trait), so that ASR/VAD/diarization consumers do not link it? We are on 0.15.4 for that reason and would like to pick up the 0.15.6 diarizer fixes (`fix(offline-diarizer): pyannote-parity clustering ...`), which measurably help the forced-speaker-count path on our fixtures.

Environment: macOS 26.5, Xcode 26, Swift 6, arm64; FluidAudio 0.15.6 tag; consumer links its own Rust staticlib built with the stable toolchain.

### LINKED ISSUE #888: NemoTextProcessing is an unconditional dependency of the FluidAudio target (~18 MB for ASR-only consumers)
We use FluidAudio for streaming ASR (Nemotron) and speaker diarization on iOS and macOS. We don't use any TTS backend or the ITN/text-normalization API.

Since v0.15.6, `NemoTextProcessing` is a dependency of the main `FluidAudio` target, so every consumer resolves and links it. #867 then replaced the `dlopen(nil)` discovery with a direct link, which seems to be why it no longer dead-strips for consumers that never call it.

What we measured going from v0.15.5 to 6428e291, same app version, universal macOS binary:

- executable 108.2 MB to 126.1 MB
- `nm | grep -c text_processing_rs`: 0 before, 650 after
- 12 of the archive's 17 exported `nemo_*` symbols present in the final binary
- the xcframework zip itself is ~49 MB, 29 MB per iOS slice

Only `TTS/Shared/NemoTextNormalizer.swift` references the normalizer. Nothing under `Sources/FluidAudio/ASR/`, `Diarizer/`, `Speaker/` or `Shared/` does, so the code is reachable only through TTS but ships regardless.

Would you consider making it opt-in? A few options, roughly in order of how invasive they look from the outside:

1. A separate product (e.g. `FluidAudioTTS`) that carries the binary dependency, leaving `FluidAudio` ASR/diarization-only.
2. A SwiftPM trait, now that 6.1 has them.
3. Restoring lazy `dlopen` for consumers that don't link the TTS path, if #867's motivation was mainly reliability rather than the linkage itself.

Happy to test a branch on a real universal build if that's useful. We build both slices and can report before/after sizes and symbol counts.

Alex-Wengg:
Closes #888, closes #880.

## Problem

Since 0.15.5 every consumer links `NemoTextProcessing`, a ~18 MB-per-slice prebuilt Rust staticlib that only the TTS frontends and the ITN API call. #888 measured +18 MB on a universal macOS binary for an ASR/diarization-only app. #880 can't link at all: a second Rust runtime duplicates `_rust_eh_personality` and 143 std symbols.

## Approach: a package trait, not a module split

Splitting TTS/ITN into a separate target would force every shared internal (`ModelHub`, `AppLogger`, download/audio utils) public and change every TTS consumer's imports. A SwiftPM trait keeps one module and one API:

- **`Package@swift-6.2.swift`** — same manifest plus a default-on `NemoTextProcessing` trait; the binary target becomes a conditional dependency. Gated at 6.2, not 6.1: SwiftPM 6.1 (Xcode 16.4) accepts `--disable-default-traits` but still links the trait-conditioned binary target (first CI run: 1944 symbols); Swift 6.2 links 0. Consumers opt out with:
  ```swift
  .package(url: "https://github.com/FluidInference/FluidAudio.git", from: "0.15.7", traits: [])
  ```
  `Package.swift` stays at tools 6.0 (CI's Xcode 16 iOS job, older consumers) and always links the engine.
- **`TextNormalizer` / `NemoTextNormalizer` stay public on every build.** Only the import and the FFI call bodies are guarded with `canImport(CNemoTextProcessing)`, which is true under both manifests whenever the engine is a dependency. Trait off: `isNativeAvailable`, `isTnAvailable`, and the new `NemoTextNormalizer.isAvailable` report `false`, every call passes text through, `version` is `nil`, rule mutations log a warning. This is the explicit-availability contract, not the silent dlopen no-op #867 removed.
- **Kokoro Mandarin** verbalizes numeric-only input with `MandarinNumberNormalizer` when the engine is absent, so `$5.50` still reaches G2P instead of the bopomofo passthrough.
- **Tests**: engine-backed classes skip when the trait is off; new `TextNormalizerUnavailableTests` / `NemoTextNormalizerUnavailableTests` pin the passthrough contract and run only then.
- **CI**: new macos-15 job on Xcode 26.3 builds `fluidaudiocli` with `--disable-default-traits`, asserts zero `text_processing_rs` / `nemo_` symbols in the binary, and runs the normalizer tests trait-off.
- **Docs**: PostProcessing.md "Opting out of the engine" section, README pointer.

## Verification (local, Swift 6.2.3)

Universal release build of `fluidaudiocli` (arm64 + x86_64 slices built per-triple, `lipo -create`), `strip -x` applied:

| build | arm64 slice | x86_64 slice | universal | engine symbols/slice |
|---|---|---|---|---|
| default | 16.15 MB | 16.90 MB | 33.05 MB | 671 / 680 |
| `traits: []` | 7.98 MB | 8.75 MB | 16.76 MB | 0 / 0 |

Delta: **-16.3 MB universal (-8.2 MB per slice)**, the same order as the +18 MB @JulianPscheid measured across two releases on a universal app in #888. Library + CLI compile clean both ways; `swift format lint` clean. XCTest is unavailable locally, so the test target is exercised by CI.

## Notes

- Relates to #890 (tvOS): the same `canImport` guards are what that PR's review asked for, so it can reduce to the `.when(platforms:)` condition on top of this.
- @JulianPscheid offered a before/after universal build in #888; a run of your app against this branch with `traits: []` would confirm the size delta on a real target.


JulianPscheid:
Ran the before/after I offered in #888. Universal release build of our macOS app, Xcode 26.3 / Swift 6.2.4, same tree and same app version for both, PR head `97c6e82`, only the `traits:` line differs.

| build | executable | arm64 slice | x86_64 slice | `text_processing_rs` (arm64 / x86_64) | `_nemo_*` per slice |
|---|---|---|---|---|---|
| default traits | 132.41 MB | 65,231,664 | 67,146,432 | 650 / 658 | 12 |
| `traits: []` | 115.56 MB | 56,868,880 | 58,666,960 | 0 / 0 | 0 |

That's 16.85 MB off the universal executable, 8.36 MB off arm64 and 8.48 MB off x86_64, or 12.7% of the binary. The `.app` bundle drops the same 16.8 MB (650.6 to 633.8 MB). Same order as your `fluidaudiocli` numbers. These are the as-built release binaries with no extra `strip -x`, so they run a bit larger than yours.

Nothing regressed on the ASR side. `AsrManager`, `StreamingAsrManager` and `DiarizerManager` symbols are all still in both slices, and we don't reference `TextNormalizer` anywhere.

Two things worth knowing, neither a blocker.

Xcode projects can't set traits. We consume FluidAudio as an `XCRemoteSwiftPackageReference` in an `.xcodeproj`, and Xcode 26.3 has nowhere to put `traits: []`: no UI for it, no pbxproj key. Trait support seems to have landed in 26.4, which I haven't tested. What worked was a one-target local package that declares the dependency and re-exports the module, with the app target linking that instead of FluidAudio directly:

```swift
// swift-tools-version: 6.2
let package = Package(
    name: "FluidAudioShim",
    platforms: [.macOS(.v14), .iOS(.v17)],
    products: [.library(name: "FluidAudioShim", targets: ["FluidAudioShim"])],
    dependencies: [
        .package(url: "https://github.com/FluidInference/FluidAudio.git", from: "0.15.7", traits: [])
    ],
    targets: [
        .target(name: "FluidAudioShim", dependencies: [.product(name: "FluidAudio", package: "FluidAudio")])
    ]
)
```

with `@_exported import FluidAudio` as the target's only source file, so existing `import FluidAudio` lines keep compiling. Might be worth a couple of lines in the "Opting out" section, since app consumers are more likely to be Xcode projects than packages.

The xcframework still downloads. With a cold `--cache-path`, `traits: []` still fetches and extracts the 49 MB `NemoTextProcessing.xcframework.zip`. The binary target is declared unconditionally and only the dependency edge is trait-conditioned, so resolution pulls it either way. Ship size is fixed, but CI and clean checkouts still pay for the download. Probably a SwiftPM limitation rather than something this PR can address.

One caveat on our end: this is a build and symbol check, not a live session run. The link succeeded and our ASR path never touched the engine, so I'd be surprised by a runtime difference, but I haven't proven one way or the other yet.

PR #895, opened 2026-09-09
fix(asr/streaming): re-decode the final window with a fresh decoder state (#855)

Alex-Wengg:
Fixes the trailing-transcript loss reported in the #855 follow-up (three public recordings, 500-clip corpus regression). Bisected to #861 itself.

## What was happening

#861 made the final streaming window decode from frame 0 with an emission cutoff, which fixed the short-flush-window tail loss (repro 1) and the repro-2 seam phantom. But it re-decoded with the **carried** decoder state. On a long final window that state has already consumed the overlap audio, and re-walking it leaves the decoder emitting blanks for the rest of the window. Instrumented on the reporter's clip 01:

```
window 1: 13.0 s, 48 tokens, last frame 159
window 2 (final): offset 112, 12.4 s, plan = (frame 0, cutoff local 42) → rawTokens = 0
```

Zero emissions for 12.4 s of audio, 9 s of which no earlier window had seen. Batch decode of the same audio is complete. Same on all three recordings, and on every revision from c34a5443 (#861) forward; absent at df1417ce.

## Fix

Start the frame-0 re-decode from a fresh `TdtDecoderState`, exactly as the batch chunker does for every chunk. Suppression and dedup are unchanged; the 2 s left context is enough for the fresh state to re-establish itself before the cutoff. Four lines plus comments.

## Verification (release build, M5 Pro, `--model-version v3 --streaming`)

| recording | main `9dfb81b9` | this branch |
|---|---|---|
| 01-validation-request-21.4s | stops at "validate" | complete, ends "help them out." |
| 02-release-readiness-19.8s | stops at "is it's" | complete, ends "cutting a release?" |
| 03-diff-explanation-16.9s | stops at "and an" | complete, ends "in that difference." |
| issue-855-repro2-17.2s (the #861 case) | 5 occurrences | 5 occurrences (batch = 5) |

So the #861 fix is kept, not reverted. Known residue on clip 03: a word cut at the previous window's edge survives as a fragment ("and an, and analyzing"); dedup matches whole tokens.

## Tests

`SlidingWindowFinalWindowRegressionTests` feeds the three recordings (cleared for public release by the speaker, per the release README) through `SlidingWindowAsrManager` in 1 s buffers and asserts `finish()` keeps each recording's final words. It runs when the v3 models are cached or `FLUIDAUDIO_RUN_ASR_E2E=1`; the ASR benchmark workflow now runs it after caching the models, so this class of regression fails CI. Docs: `LongTranscription.md` "Streaming Final Window".

## Corpus confirmation

@saurabhav88 ran the branch through a frozen 500-recording corpus (three arms, two full passes each, byte-identical): 489/500 unchanged, every one of the 11 changed recordings is ≥ 12.98 s, changed-set distance to the batch proxy 70 → 31, five truncated tails recovered, no output shorter, batch byte-identical. Two non-blocking findings recorded there and tracked separately, not in this PR: a rare mid-word seam join on one recording (`everyturing`, same class as the clip-03 residue but a non-word), and two mid-stream duplications (`just just`, `make make`) that `main` introduced and this PR does not touch.

## Review round

- Swift 6.1 strict concurrency rejected sending a buffer that shared a region with its `AVAudioFile`; the test now decodes once to `[Float]` and builds each 1 s chunk as its own buffer before sending it to the actor.
- The Fixtures resource is declared in both manifests, so Swift 6.2+ runs don't skip the test.


saurabhav88:
Ran this patch through the 500-recording corpus as offered. Headline: we would take it. It recovers the
tail on every recording where current `main` truncates, it changes nothing on any recording shorter than
one window, and batch is untouched. Two findings that are not blockers are at the bottom, one of them a
new artifact this patch introduces.

## Setup

Three arms, all stock, all `swift build -c release --product fluidaudiocli`, Parakeet TDT v3, Apple M5 Max,
macOS 26.7, Swift 6.3.3:

| arm | revision |
|---|---|
| `base` | `df1417ce86225af50150b11eb04fa01c0aba9148` (before #861) |
| `main` | `9dfb81b9535e119f855f9ec9c98308ea6076cc95` |
| `pr895` | `cc678743ace2f91f166af7ed88c300ea5113c77a` (this PR) |

```
fluidaudiocli transcribe <clip.wav> --model-version v3 --streaming
fluidaudiocli transcribe <clip.wav> --model-version v3
```

Corpus: 500 real dictations captured by our app, each the exact 16 kHz mono Int16 buffer supplied to
FluidAudio. Frozen with a per-file SHA-256 before the runs, because the source directory rotates. 2,378 s
of audio, 4.8 s mean, 17 recordings at or above 13 s. Batch decode of the same audio is the proxy for what
was spoken. It is an ASR output and not human ground truth, so everything below is stated as a distance to
that proxy, on folded tokens (lowercased, punctuation stripped).

**Two caveats on the corpus, both ours.** It is a refreshed snapshot, not the identical 500 recordings
behind the 8-closer/30-farther split in #855; that set was not retained. And it is short-skewed, so it has
far less to say about a final-window defect than its size suggests. The three public recordings carry more
weight here than the corpus does.

**Determinism.** Every arm and both modes were run twice, complete, in separate passes: 2,500 decodes,
byte-identical across the two passes on all five result sets. No sampling, no averaging.

## The three public recordings reproduce your table exactly

| recording | `base` | `main` | `pr895` |
|---|---|---|---|
| `01-validation-request-21.4s` | complete | stops at "validate" | complete |
| `02-release-readiness-19.8s` | complete | stops at "is it's" | complete |
| `03-diff-explanation-16.9s` | complete | stops at "and an" | complete |

Your noted residue on clip 03 ("and an, and analyzing") is present on `base` too, so this patch does not
introduce it.

## Corpus result

- **489 of 500 byte-identical across all three arms.**
- **11 changed. Every one is 12.98 s or longer.** Every recording below that is untouched by both #861 and
  this patch, which is the scoping the patch claims. Seven recordings above 13 s were also unchanged,
  including a 147 s one.
- **Batch is untouched: `pr895` batch is byte-identical to `main` batch on all 500.**

Per recording, folded-token distance to the batch proxy:

| clip | duration | words base/main/pr895 | distance base/main/pr895 |
|---|---|---|---|
| `08CBD384` | 46.2s | 104/104/109 | 15/15/**12** |
| `63CE980C` | 24.6s | 41/27/41 | 0/14/**0** |
| `7D9241C2` | 23.8s | 49/55/55 | 7/1/1 |
| `3E9FACAD` | 22.4s | 38/39/39 | 0/1/1 |
| `99C654B5` | 20.0s | 50/34/51 | 1/17/**0** |
| `965854B8` | 17.9s | 38/39/39 | 3/4/4 |
| `76CB773B` | 16.6s | 39/32/39 | 1/6/**1** |
| `3B8A7579` | 16.6s | 30/30/30 | 0/0/0 |
| `12D5CDD1` | 16.0s | 41/40/41 | 0/1/**0** |
| `D67BD36D` | 13.4s | 27/25/27 | 2/1/2 |
| `0A575FDA` | 13.0s | 32/32/34 | 10/10/10 |
| **total** | | | **39/70/31** |

### The regressed set

Six recordings where `main` sits farther from the batch proxy than `base`. This patch restores two exactly,
improves two, leaves two unchanged, and worsens none.

| clip | base | main | pr895 | verdict |
|---|---|---|---|---|
| `99C654B5` | 1 | 17 | 0 | better than base; recovers a word base also lost |
| `63CE980C` | 0 | 14 | 0 | restored to base exactly |
| `76CB773B` | 1 | 6 | 1 | restored to base exactly |
| `12D5CDD1` | 0 | 1 | 0 | better than base |
| `3E9FACAD` | 0 | 1 | 1 | unchanged by the patch |
| `965854B8` | 3 | 4 | 4 | unchanged by the patch |

Shape check, independent of the proxy: `main` truncates `base` (its folded tokens are a strict prefix) on
four recordings, dropping 16, 14, 7 and 2 words. `pr895` extends `main` on five, adding 17, 14, 7, 5 and 2
words. Nothing in the corpus loses words on this patch.

**The patch beats the pre-#861 baseline on one recording.** On the 46.2 s clip, `base` also drops the final
five words. `pr895` is the only arm that keeps them. That is worth knowing: a fresh state on the final
window is not merely a revert of #861's side effect.

**And `main` is not uniformly worse.** On `7D9241C2` (23.8 s) `base` drops the closing question and both
`main` and `pr895` keep it, 7 to 1.

## Two findings

**1. New artifact, introduced by this patch.** On `0A575FDA` (13.0 s) the recovered tail joins mid-word and
produces a token nobody said:

```
base  : ...but it's still capture everyek.
main  : ...but it's still capture every
pr895 : ...but it's still capture everyturing everything saying.
```

All three arms are wrong on this recording and the distance is 10 in each, so it does not move the
totals. It is the same class as your clip-03 residue, with one difference worth weighing: whole-token dedup
lets a real word survive twice ("and an, and analyzing"), while here the join manufactures `everyturing`,
which is not a word and is not recoverable downstream. One recording out of 500, and out of five where the
patch recovers a tail. Not a blocker for us; flagging it because a nonword is a different failure than a
duplicated word.

**2. Not addressed by this patch, and still open from #855.** Two recordings carry a mid-stream word
duplication that `main` introduced and this patch leaves in place: `3E9FACAD` ("about just just speaking",
base has one "just") and `965854B8` ("we need to m make make that work", base has one "make"). These are
seam duplications rather than final-window losses, so they look like the dedup half of #855 rather than
anything this PR set out to fix. No action requested here; recording it so the #855 close-out does not read
as covering them.

**One metric note, so the table is not read wrongly.** On `D67BD36D` the distance says `pr895` is farther
than `main` (2 against 1). The content says otherwise: `main` drops the final word entirely, `pr895`
recovers it and misspells a product name that the batch proxy also gets wrong. A dropped word costs one
edit and a two-word misspelling costs two, so the proxy penalises the arm that recovered the audio. That is
the only recording where `pr895` scores farther than `main` anywhere in the 500.

## Verdict

On our data this patch is a strict improvement: 70 to 31 on total distance across the changed set, tails
recovered on five recordings, one of them beyond what the pre-#861 baseline managed, nothing made shorter,
batch unaffected, and the change confined to sessions long enough to have a final window. We would ship it.

Happy to rerun any specific arm, or to publish the two recordings behind the seam artifact if that would
help.

Issue #897, opened 2026-09-09
Streaming: residual seam artifacts after #895 (mid-word join, mid-stream word duplication)


Alex-Wengg:
Split out of #855 so its close-out (fixed by #861 + #895) does not read as covering these. Both come from @saurabhav88's 500-recording corpus run on #895 (https://github.com/FluidInference/FluidAudio/pull/895#issuecomment, "Two findings").

## 1. Mid-word join at the final-window seam (new class, non-word output)

On one 13.0 s recording the recovered tail joins mid-word and manufactures a token nobody said:

```
base  : ...but it's still capture everyek.
pr895 : ...but it's still capture everyturing everything saying.
```

Same class as the documented clip-03 residue (`and an, and analyzing`: a word cut at the previous window's edge survives as a fragment because dedup matches whole tokens), but here the join produces a non-word, which is not recoverable downstream. 1/500 recordings, 1/5 of the recovered tails. Candidate fix: when the previous window's last token is a word fragment (no trailing space / sub-word) and the re-decoded window's first post-cutoff token begins within the jitter margin, drop the fragment rather than the re-decoded token.

## 2. Mid-stream word duplication (pre-existing, introduced between df1417ce and 6428e291, not touched by #895)

Two recordings carry a duplicated word at an interior seam that `df1417ce` did not have:

- `about just just speaking` (base: one `just`)
- `we need to m make make that work` (base: one `make`)

These are interior-window seams, so they are the dedup half of #855, not the final-window path #861/#895 changed. Not yet bisected; candidates are the same window of commits (c34a5443, 1bed3e21, 0dc46a3c, c92e4aea) minus the final-window plan, which only fires on `isLastChunk`.

## Reproduction

The corpus is private; the reporter has offered to run targeted patches through it and report per-recording deltas. A public fixture for either artifact would let `SlidingWindowFinalWindowRegressionTests` (Tests/FluidAudioTests/ASR/Parakeet/SlidingWindow/Fixtures) cover it.

Alex-Wengg:
[...]

PR #903, opened 2026-09-10
fix(asr/streaming): reconcile the final-window seam — retire the previous window's edge-cut word (#897)

Alex-Wengg:
Addresses the mid-word join class in #897 (the first of its two findings). The interior-seam duplications (`just just` / `make make`) are a different mechanism and stay open there.

## Mechanism (clip 03 of the #855 fixtures, token trace)

Window 1 ends `▁and`@153 `▁an`@156 — the fragment of "analyzing", cut by the window edge. The final window's cutoff was anchored at that fragment's frame, so the re-decode emitted `,`@152 `▁and`@155 `▁anal`@159 `y` `z` `ing` after it. Dedup could not repair the join: the fragment never equals the full word, a re-emitted single token is below the substring matcher's two-token minimum, and the comma the decoder attaches at its emission start blocks the suffix–prefix match. Output: `code and an, and analyzing`. The reporter's corpus shows the same class yielding a non-word (`everyturing`).

## Fix

- The cutoff anchors at the previous window's **last word start**, so the final window re-decodes that word in full.
- `reconcileFinalWindowSeam` (pure) retires the previous last word from the accumulated output and strips the re-decode's seam artifacts: punctuation emitted at or before that word's frame (classified by piece text; `ASRConstants.punctuationTokens` lists only sentence-final marks and the artifact is a comma), and tokens duplicating a kept previous token inside the jitter region.
- `transcribeChunk` reports `droppedPreviousTokens`; `SlidingWindowAsrManager` retires them from the token arrays and trims the word from the volatile/confirmed text (whole-word suffix only), so both the token path and the boosting text path see the reconciled seam.

## Verification (release, M5 Pro)

| clip | before | after |
|---|---|---|
| 03 streaming | `code and an, and analyzing` | `code and analyzing` (= batch) |
| 01 / 02 streaming | complete | unchanged |
| repro-2 (#861 case) | 5 / 5 | 5 / 5 |
| earnings 9 s / 15 s, boosted streaming | corrected | unchanged |
| clip 01, boosted, no confirmed window | `follow-up`, tail intact | unchanged |

## Review round (CI)

The first revision passed the clip-03 fixture assertion on CI but the boosting e2e (1 s buffers, which yields an empty 0.4 s flush window) lost its final word: the previous window's `out.` was retired on the promise of a re-decode that emitted nothing. Two guards, both pinned by unit tests on the token shapes:

- retire the previous last word only when the re-decode has a token past the previous window's last frame (plus jitter), the only case in which that word can be an edge-cut fragment; an empty or early-ending final window keeps it;
- if the re-decode's first word is the same word the previous window ended on, it was complete: keep the previous copy (it carries the sentence-final punctuation a re-decode at the audio end tends to omit) and drop the re-emission as a duplicate. Only a different first word (`an` vs `analyzing`) retires it.

Re-verified across chunk sizes 7 / 10 / 11, boosted and unboosted: clip 03 still equals batch at the seam, every tail intact, repro-2 5/5, earnings corrections unchanged. (At chunk 10 the interior window itself decodes `help them out` without a period and the flush window is empty; that is the decoder's output, identical on `main`.)

Two gotchas hit on the way, both now covered by tests: the loaded vocabulary stores word-initial pieces with a leading space rather than the SentencePiece marker (the batch chunker already checks both), and the punctuation id constant has no comma.

Tests: plan anchoring, `trailingWordStartIndex` (both boundary forms), `isPunctuationPiece`, reconciliation on the clip-03 token shape and on genuine leading tokens, `removingTrailingWord`; the real-recording fixture test asserts clip 03 contains `code and analyzing` and not `and an,`, and runs in the ASR benchmark job.

@saurabhav88 if the corpus is still frozen, the `0A575FDA` recording (`everyturing`) is the one to re-run on this branch.

### DIFF
--- a/Documentation/ASR/LongTranscription.md
+++ b/Documentation/ASR/LongTranscription.md
@@ -423,8 +423,19 @@ that dedup strips. Two rules make this safe:
   exceed dedup's bounded search; suppressing at the source keeps dedup's job
   to the jitter margin.
 
-Known residue: a word cut by the previous window's edge (`and an` → `and
-analyzing`) can survive as a fragment, since dedup matches whole tokens.
+**Seam reconciliation (#897).** The previous window's last word may be a
+fragment cut by the window edge (`and an` for `and analyzing`), which dedup
+can never repair: the fragment never equals the full word, a re-emitted single
+token is below the substring matcher's two-token minimum, and a punctuation
+token the decoder attaches at its emission start blocks the suffix–prefix
+match. So the cutoff anchors at the previous window's *last word start*
+(`lastWordStartFrame`), the final window re-decodes that word in full, and
+`AsrManager.reconcileFinalWindowSeam` retires the previous last word from the
+accumulated tokens and text (`droppedPreviousTokens`) while stripping the
+re-decode's seam artifacts: punctuation emitted at or before that word's
+frame, and tokens duplicating a kept previous token inside the jitter region.
+Clip 03 of the fixtures pins it: `code and an, and analyzing` → `code and
+analyzing`, matching batch.
 
 Regression fixtures: `Tests/FluidAudioTests/ASR/Parakeet/SlidingWindow/Fixtures`
 (three real recordings, cleared for release by the speaker), exercised by
--- a/Sources/FluidAudio/ASR/Parakeet/SlidingWindow/SlidingWindowAsrManager.swift
+++ b/Sources/FluidAudio/ASR/Parakeet/SlidingWindow/SlidingWindowAsrManager.swift
@@ -445,7 +445,22 @@ public actor SlidingWindowAsrManager {
             // Update stored decoder state
             self.decoderState = state
 
-            let (tokens, timestamps, confidences, _) = result
+            let (tokens, timestamps, confidences, _, droppedPreviousTokens) = result
+
+            // Final window re-decoded the previous window's last word in full (#897):
+            // retire that word from the accumulated tokens and from the text state.
+            if droppedPreviousTokens > 0, droppedPreviousTokens < accumulatedTokens.count {
+                let dropped = Array(accumulatedTokens.suffix(droppedPreviousTokens))
+                accumulatedTokens.removeLast(droppedPreviousTokens)
+                accumulatedTokenTimestamps.removeLast(min(droppedPreviousTokens, accumulatedTokenTimestamps.count))
+                if let droppedText = await asrManager?.convertTokensToText(dropped), !droppedText.isEmpty {
+                    if let trimmed = Self.removingTrailingWord(droppedText, from: volatileTranscript) {
+                        volatileTranscript = trimmed
+                    } else if let trimmed = Self.removingTrailingWord(droppedText, from: confirmedTranscript) {
+                        confirmedTranscript = trimmed
+                    }
+                }
+            }
 
             let adjustedTimestamps = Self.applyGlobalFrameOffset(
                 to: timestamps,
@@ -590,6 +605,14 @@ public actor SlidingWindowAsrManager {
                 ? "insufficient context (\(String(format: "%.1f", totalAudioProcessed))s)" : "low confidence"
             logger.debug("VOLATILE (\(result.confidence)): \(reason) - updated volatile '\(result.text)'")
         }
+    }
+
+    /// `text` without its trailing `word` when `text` ends with that word as a
+    /// whole word (equal, or preceded by a space); nil otherwise. Pure.
+    static func removingTrailingWord(_ word: String, from text: String) -> String? {
+        if text == word { return "" }
+        guard text.hasSuffix(" " + word) else { return nil }
+        return String(text.dropLast(word.count + 1))
     }
 
     /// Join the still-volatile text with a newer unconfirmed window's text.
--- a/Sources/FluidAudio/ASR/Parakeet/SlidingWindow/TDT/AsrManager+Transcription.swift
+++ b/Sources/FluidAudio/ASR/Parakeet/SlidingWindow/TDT/AsrManager+Transcription.swift
@@ -71,14 +71,105 @@ extension AsrManager {
         isLastChunk: Bool,
         previousTokens: [Int],
         previousTokenTimestamps: [Int]?,
-        globalFrameOffset: Int
+        globalFrameOffset: Int,
+        lastWordStartFrame: Int? = nil
     ) -> (initialTimeIndexOverride: Int?, emitTokensAfterFrame: Int?) {
         guard isLastChunk, let previousTimestamps = previousTokenTimestamps, !previousTokens.isEmpty else {
             return (nil, nil)
         }
-        let lastEmittedGlobalFrame = previousTimestamps.max() ?? 0
-        let cutoff = max(0, lastEmittedGlobalFrame - globalFrameOffset - redecodeEmissionJitterFrames)
+        // Anchor the cutoff at the previous window's last *word*, not its last
+        // token: that word may have been cut by the window edge (#897), and the
+        // re-decode must be free to re-emit it in full.
+        let anchorGlobalFrame = lastWordStartFrame ?? (previousTimestamps.max() ?? 0)
+        let cutoff = max(0, anchorGlobalFrame - globalFrameOffset - redecodeEmissionJitterFrames)
         return (0, cutoff)
+    }
+
+    /// Index of the first token of the previous window's last word — a piece
+    /// carrying the SentencePiece word boundary, or the leading space the loaded
+    /// vocabulary normalizes it to — or nil when the sequence has no word start
+    /// after its first token (dropping index 0 would discard the whole window).
+    /// Pure, for testability.
+    nonisolated internal static func trailingWordStartIndex(pieces: [String]) -> Int? {
+        guard
+            let idx = pieces.lastIndex(where: {
+                $0.hasPrefix(ASRConstants.sentencePieceWordBoundary) || $0.hasPrefix(" ")
+            }), idx > 0
+        else { return nil }
+        return idx
+    }
+
+    /// A vocabulary piece that is nothing but punctuation once the word
+    /// boundary (marker or normalized leading space) is stripped.
+    nonisolated internal static func isPunctuationPiece(_ piece: String) -> Bool {
+        var core = piece
+        if core.hasPrefix(ASRConstants.sentencePieceWordBoundary) {
+            core.removeFirst(ASRConstants.sentencePieceWordBoundary.count)
+        }
+        core = core.trimmingCharacters(in: .whitespaces)
+        guard !core.isEmpty else { return false }
+        return core.unicodeScalars.allSatisfy { CharacterSet.punctuationCharacters.contains($0) }
+    }
+
+    /// Seam reconciliation for the final streaming window (#897).
+    ///
+    /// The previous window's trailing word may be a fragment cut by the window
+    /// edge (`and an` for `and analyzing`); the re-decoded final window emits
+    /// that word in full from `trailingWordStart`'s frame minus the jitter
+    /// margin. Dedup cannot fix this — the fragment never equals the full word,
+    /// a re-emitted single token is below the substring matcher's minimum run,
+    /// and a boundary punctuation the decoder attaches at its emission start
+    /// blocks the suffix–prefix match.
+    ///
+    /// Returns how many trailing previous tokens to drop (the whole last word)
+    /// and how many leading current tokens to drop: punctuation emitted at or
+    /// before the previous last word's frame (a seam artifact), and tokens that
+    /// duplicate a kept previous token within `frameTolerance` inside the jitter
+    /// region. Pure, for testability; timestamps are global frames.
+    nonisolated internal static func reconcileFinalWindowSeam(
+        previousTokens: [Int],
+        previousTimestamps: [Int],
+        trailingWordStart: Int,
+        currentTokens: [Int],
+        currentTimestamps: [Int],
+        currentPieces: [String] = [],
+        punctuationTokens: [Int] = ASRConstants.punctuationTokens,
+        jitterFrames: Int = redecodeEmissionJitterFrames,
+        frameTolerance: Int = ASRConstants.duplicateFrameTolerance
+    ) -> (droppedPrevious: Int, droppedCurrent: Int) {
+        guard trailingWordStart > 0, trailingWordStart < previousTokens.count,
+            previousTimestamps.count == previousTokens.count,
+            currentTimestamps.count == currentTokens.count
+        else { return (0, 0) }
+
+        // `ASRConstants.punctuationTokens` lists only sentence-final marks; the
+        // seam artifact is usually a comma, so classify by the piece text too.
+        func isPunctuation(_ index: Int) -> Bool {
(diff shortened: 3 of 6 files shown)


saurabhav88:
Ran `0A575FDA` on this branch as asked, then ran a fresh 500-recording corpus on it as well. Three findings. The second is the one I would treat as a blocker: this branch loses a real trailing word on a recording already in your own test suite, and `main` does not.

## Setup

| arm | revision |
|---|---|
| `base` | `df1417ce86225af50150b11eb04fa01c0aba9148` (before #861) |
| `main` | `600d94e52a4cb31e4997cf99ea2c86600270579d` |
| `pr903` | `0043569a6c2bb12b7a12a0f065d8e45d9a3eccf1` (this branch) |

Apple M5 Max, macOS 26.7, Swift 6.3.3, Parakeet TDT v3, `swift build -c release` per arm.

Two ways of driving it. Single recordings through `fluidaudiocli transcribe … --streaming` and `… --model-version v3`. The corpus through a small runner that loads `AsrModels` once and, per file, runs a fresh `AsrManager` + `TdtDecoderState` for the one-shot path and a fresh `SlidingWindowAsrManager` for the streaming path, fed in 11 s chunks with the same 0.5 s settle before `finish()` that the CLI uses. Control: the runner reproduces the CLI's output byte-for-byte on `0A575FDA` in both modes. Provenance, since the arms differ by one commit: `nm -a` finds `trailingWordStartIndex` in the `pr903` binary and not in the `main` one.

**Corpus caveat, ours.** The frozen 500 behind the #895 report was not retained — that scratch directory is gone. This is a fresh snapshot of 500 recordings from our app's own store: 2,032 s of audio, mean 4.1 s, **median 2.9 s, and only 15 recordings at or above 13 s**. It is short-skewed, so it has far less to say about a final-window defect than 500 suggests. Every arm ran twice in separate passes, byte-identical both times.

## Finding 1 — the mid-word join survives on `0A575FDA`, in a different shape

One-shot decode of the same audio, identical on all three arms, as the proxy for what was said:

> ...but it's still **capturing everything** in the same.

Streaming:

| arm | tail |
|---|---|
| `base` | ...still **capture everyek.** |
| `main` | ...still **capture everyturing everything** saying. |
| `pr903` | ...still **captureturing everything** saying. |

The branch removes the duplicated `everything`. The non-word survives; it moves from `everyturing` to `captureturing`.

Debug trace on this branch:

```
Chunk 1: '... but it's still capture every'
Final-window seam: dropped 2 trailing previous token(s), 1 leading current token(s)
Chunk 2: 'turing everything saying.'
```

Window 1 ends on `every`, the edge-cut fragment. The branch retires it, correctly. But the final window's own re-decode **begins inside the earlier word** `capturing`, and its first emitted piece is `turing` — a word-tail fragment, not a word start. Retiring the previous window's last word does nothing about that, so `capture` + `turing` glue into `captureturing`.

I first assumed the cutoff anchor was landing late and the re-decode was losing the front of its own word. That is falsifiable and false: `redecodeEmissionJitterFrames` raised 5 → 40 changes `droppedCurrent` from 1 to 0 and leaves `Chunk 2` byte-identical at `turing everything saying.`. The leading fragment is not being suppressed by the cutoff — it is emitted and kept.

## Finding 2 — the branch loses a trailing word that `main` keeps

`Tests/…/SlidingWindowVocabularyBoostingStreamingTests.swift:101`, `testInMemoryTermsRescoreAndKeepTailOnRealRecording` — the end-to-end test #898 added on `01-validation-request-21.4s`.

```
swift test --filter "SlidingWindowVocabularyBoostingStreamingTests"
```

| arm | result |
|---|---|
| `main` | **passed**, 2 runs (1.838 s / 1.060 s) |
| `pr903` | **failed**, 2 runs, identical assertion |

```
XCTAssertTrue failed - tail lost: Hey, before we go to them … let's craft a polite follow-up for them to help them
```

`out` is gone. Trace:

```
Final-window seam: dropped 1 trailing previous token(s), 0 leading current token(s)
Chunk 3: ''
```

The final window emitted nothing, and the previous window's last word had already been retired on the strength of a re-decode that then produced no replacement. `droppedPrevious` is taken from `seam.droppedPrevious` before anything checks that the re-decode returned usable tokens, so an empty final window deletes a correct word. Silent text loss rather than an artifact, which is why I would hold on this one rather than on finding 1.

## Finding 3 — on 500 real dictations the branch is net level, and breaks two clean recordings

Word-level edit distance of each streaming arm to the one-shot decode of the same audio, folded (lowercased, punctuation stripped), summed over all 500:

| arm | distance | of 4,194 reference words |
|---|---|---|
| `main` | 47 | 1.12% |
| `pr903` | 47 | 1.12% |

**7 of 500 recordings change at all: 3 closer to the one-shot decode, 2 further, 2 level.**

Closer:

| id | `main` | `pr903` |
|---|---|---|
| `3E9FACAD` | `about just just speaking` | `about just speaking` |
| `965854B8` | `need to m make make that work` | `need to m make that work` |
| `0A575FDA` | `capture everyturing everything saying` | `captureturing everything saying` |

Further:

| id | `main` | `pr903` |
|---|---|---|
| `99C654B5` | `the words appear in the box` — **exact match to the one-shot decode** | `the words appear in thex` |
| `D28C4EBC` | `and it does it properly` (one-shot: `properly`) | `and it does it correctly` |

`99C654B5` is the one I would look at first: distance 0 → 2 on a recording `main` gets exactly right, and the failure is the same shape as finding 1 — a kept word plus a re-decoded word-tail fragment, glued. `D28C4EBC` is not a seam artifact but a decode that moved: changing what the re-decode is allowed to emit changed a word 4 s earlier in the same final window.

Level, both punctuation-only and both mild improvements the folded metric cannot see (`sentence or two. in the` → `sentence or two in the`, `build scripts. around it` → `build scripts around it`): `08CBD384`, `5390B90D`.

Counting only invented words — a streaming token absent from the one-shot decode of the same audio that no dictionary knows — `main` has 5 across the corpus and `pr903` has 6. The extra one is `thex`.

## Two repairs, verified locally

Both against `0043569a`, in `reconcileFinalWindowSeam` and its call site. Yours to take, modify or reject.

```swift
// In reconcileFinalWindowSeam, ahead of the punctuation clause: a leading piece
// that does not start a word is the tail of a word the window boundary cut, and
// that word was already emitted by the previous window (#897).
func isWordStart(_ index: Int) -> Bool {
    guard index < currentPieces.count else { return true }
    let piece = currentPieces[index]
    return piece.hasPrefix(ASRConstants.sentencePieceWordBoundary) || piece.hasPrefix(" ")
}

var droppedCurrent = 0
for (index, (id, frame)) in zip(currentTokens, currentTimestamps).enumerated() {
    if !isWordStart(index) {
        droppedCurrent += 1
        continue
    }
    // ... existing punctuation and duplicate clauses unchanged
```

```swift
// At the call site in transcribeChunk: retiring the previous window's last word
// is only safe once the re-decode has actually produced a replacement.
if seam.droppedCurrent > 0 {
    currentTokens.removeFirst(seam.droppedCurrent)
    currentTimestamps.removeFirst(seam.droppedCurrent)
    currentConfidences.removeFirst(min(seam.droppedCurrent, currentConfidences.count))
}
if currentTokens.isEmpty {
    droppedPrevious = 0
} else {
    droppedPrevious = seam.droppedPrevious
    effectivePrevious = Array(previousTokens.prefix(trailingWordStart))
    effectivePreviousTimestamps = Array(previousTimestamps.prefix(trailingWordStart))
}
```

Suggestion on the second: the emptiness rule reads better inside `reconcileFinalWindowSeam` itself — return `(0, 0)` when `currentTokens` is empty or when every current token was dropped — because it is then a property of the pure function and a unit test can pin it without models. I left it at the call site only to keep the probe minimal.

With both applied:

| check | `pr903` | `pr903` + both repairs |
|---|---|---|
| `0A575FDA` streaming | `capture` **`turing`** `everything saying` | `capture everything saying`, no non-word |
| `testInMemoryTermsRescoreAndKeepTailOnRealRecording` | failed, `out` lost | passed |
| clips 01 / 02 / 03 streaming | — | byte-identical to `pr903`; clip 03 still `code and analyzing` |
| `swift test --filter "SlidingWindow\|TokenDeduplication"` | 1 failure | 73 tests, 5 skipped, 0 failures |

## Limits, stated

- The repairs are verified on four recordings (`0A575FDA` plus clips 01–03) and the suite above. I have **not** re-run the 500 corpus with them applied, so I am not claiming a corpus-level improvement for them — only that they fix the two failures named and disturb nothing in the fixtures.
- `capture` for `capturing` on `0A575FDA` comes out of window 1 on every arm including `base`. It is a mis-hear, not a seam artifact, and neither repair addresses it.
- The one-shot decode is an engine output, not human ground truth. Every figure above is a distance to it, never an accuracy.
- A leading continuation piece is a fragment of an already-emitted word *by construction* in the final-window case, but I have not proved that holds for every vocabulary the loader can produce — the normalized-leading-space form your `trailingWordStartIndex` already handles is why the check tests both prefixes.
- No WER benchmark behind any of this. Your ASR job is the number I cannot produce here.

Happy to re-run the corpus on a revised branch, including with the repairs applied, if that would help.

saurabhav88:
Ran the 500-recording corpus with the repairs applied, which my last comment said I had not done. **Correction to that comment: do not take the first repair as I posted it — it deletes a word.** Full numbers below, and a narrower recommendation at the end.

Same setup as before (`main` `600d94e5`, `pr903` `0043569a`, Apple M5 Max, macOS 26.7, Swift 6.3.3, Parakeet TDT v3, release builds, one `AsrModels` load per corpus run, both passes byte-identical). Four arms, 500 recordings, 2,032 s of audio, median 2.9 s, 15 recordings at or above 13 s. The one-shot decode is byte-identical across all four arms, so it is a stable reference for the streaming comparison.

| arm | word distance to one-shot | exact matches | recordings changed vs `main` | recordings with an invented word | recordings with a doubled word |
|---|---|---|---|---|---|
| `main` | 47 | 487 / 500 | — | 5 | 4 |
| `pr903` | 47 | 486 / 500 | 7 | 6 | 2 |
| `pr903` + the two repairs I posted, refined | 47 | 486 / 500 | 7 | 4 | 4 |
| `pr903` + the empty-re-decode guard alone | 47 | 486 / 500 | 7 | 6 | 2 |

Distance is folded word-level edit distance summed over all 500 (lowercased, punctuation stripped); the reference is 4,194 words. Invented word = a streaming token absent from the one-shot decode of the same audio that the system dictionary does not know, after suffix stripping. Doubled = an adjacent repeat the one-shot decode does not repeat.

## The correction

**The first repair as posted deletes `box` on `99C654B5`.** That recording is one `main` gets exactly right. `pr903` turns `the words appear in the box` into `in thex`; my posted repair turns it into `in the`, which is a worse failure of the same kind you are fixing here, and the same kind as the lost `out` in my finding 2.

The cause is that two situations reach the drop identically and the leading fragment's own text is what separates them:

- `capture every` + `turing everything saying` — the fragment belongs to the **earlier** word `capturing`, and `everything` does re-emit the retired `every`. Retiring is right.
- `in the box` + `x, but she's still talking` — the fragment is the **tail of the retired word itself** and nothing re-emits it. Retiring deletes `box`.

Frames do not separate them: the seam comma sits inside the retired word's own span in both cases, so a `firstKeptFrame <= retiredEnd + jitter` test passes in both (measured, not assumed). What does separate them is that an edge-cut word is a **prefix** of the word that replaces it — `an` for `analyzing`, `every` for `everything` — which is the premise the whole reconciliation already rests on. Requiring the first surviving word to start with the retired text restores `box` and still fixes `0A575FDA`.

**That refinement is also not shippable as it stands.** It drops the invented-word count from 5 to 4, and puts the doubled-word count back from 2 to 4: `sentence or two. two in the`, `build scripts. scripts around it`, `does it properly. correctly.`. The re-emission can arrive after a punctuation token that the head scan does not drop, so a prefix test on the first surviving token alone misses it and both copies survive.

It also changes the contract of `reconcileFinalWindowSeam`, so two of your unit tests move: `testReconcileFinalWindowSeam_DropsFragmentSeamCommaAndReemittedWord` needs `previousPieces` supplied, and `testReconcileFinalWindowSeam_KeepsGenuineLeadingTokens` asserts `droppedPrevious == 1` for a current window that starts with an unrelated new word, which under the re-emission rule is `0` — keeping the word is the point. With those two updated the filtered suite is 73 tests, 5 skipped, 0 failures.

## What the corpus says about the mechanism

Every arm lands on 47. Three recordings improve and three degrade, in every variant I measured. Each version trades one artifact class for the other: `pr903` buys doubled 4 → 2 and pays one invented word; the refinement buys invented 5 → 4 and pays doubled 2 → 4.

Caveat that cuts against reading too much into this: the corpus is short-skewed, median 2.9 s, and only 15 recordings reach 13 s, so it exercises the final-window seam on a handful of recordings. It is evidence that the rework does not help ordinary dictation, not evidence that the seam class does not matter.

## Narrower recommendation

**Take the empty-re-decode guard on its own, and leave the fragment handling out.**

```swift
// In transcribeChunk, at the seam call site: retiring the previous window's last
// word is only safe once the re-decode has produced a replacement.
if seam.droppedCurrent > 0 {
    currentTokens.removeFirst(seam.droppedCurrent)
    currentTimestamps.removeFirst(seam.droppedCurrent)
    currentConfidences.removeFirst(min(seam.droppedCurrent, currentConfidences.count))
}
if currentTokens.isEmpty {
    droppedPrevious = 0
} else {
    droppedPrevious = seam.droppedPrevious
    effectivePrevious = Array(previousTokens.prefix(trailingWordStart))
    effectivePreviousTimestamps = Array(previousTimestamps.prefix(trailingWordStart))
}
```

Measured on that arm:

- `testInMemoryTermsRescoreAndKeepTailOnRealRecording` **passes**; the filtered suite is 27 tests, 0 failures.
- Output is **byte-identical to `pr903` on all 500 recordings** — zero differences. It only changes the case where the final window emits nothing, which no corpus recording hits and your 21.4 s fixture does.

So it closes finding 2 at no measured cost, and the mid-word join on `0A575FDA` stays open. I would rather leave that open than trade it for a deleted or duplicated word. Reading it inside the function is probably the better home — return `(0, 0)` when `currentTokens` is empty or when every current token was dropped — since a unit test can then pin it without models.

## Limits

- Four arms, one corpus, one machine. No WER benchmark; your ASR job is the number I cannot produce.
- The one-shot decode is an engine output, not human ground truth. Every figure above is a distance to it.
- The invented-word and doubled-word counts come from the system dictionary with crude suffix stripping, so brand names land in them; I checked each hit by hand and the counts above exclude nothing.
- I stopped after three attempts at the fragment rule rather than trying a fourth. Each fix was smaller and cleverer than the last and the distance column never moved, which is the signal I was refining the wrong thing.

Happy to run any revision through the same corpus.

Alex-Wengg:
Thanks for running it, and for the correction — you were right that the first rule deleted `box`, and right about why. The revision is https://github.com/FluidInference/FluidAudio/pull/904: continuation pieces at the re-decode's head never justify retiring, the first real word decides by text (same → keep, prefix → retire, different-and-overlapping only when the re-decode started the word itself → retire, later → keep), and head duplicates including punctuation are stripped, which should close the `two. . two` shape your prefix refinement hit. The `box` and `every` token shapes from your traces are unit tests now. Predictions for the corpus are in the PR; if the distance column still doesn't move I'll take your signal and stop.

PR #928, opened 2026-09-17
Load local Orukeet-compatible ASR bundles without repository fallback

Nathan-Roll1:

`AsrModels.load(from:)` resolves the selected NVIDIA repository even when the caller supplies a directory. This adds `loadLocal(from:)` and `--local-model-dir` so compatible Orukeet weights load from the exact compiled bundle, with missing components failing locally. The existing default download path and dependencies are preserved.

The Orukeet example consumes the pinned Hugging Face JSON manifest and verifies the archive's SHA-256 and byte count before compilation. These are normal Hugging Face model downloads; the local loader makes no network requests.

Initial validation at `0e4dda3` on an M5 Max: the library and CLI compiled, and the CLI transcribed a real recording through the new option. English, German and French recordings each produced identical text across three fresh decoder states (nine speech decodes); one-second silence returned empty text on three runs. The missing-directory check failed at the requested path without repository fallback. The changed Swift files pass formatting and whitespace checks.

The complete default package build, including NeMo text processing, also passed at `0e4dda3` on [Xcode 26.6 CI](https://github.com/Nathan-Roll1/FluidAudio/actions/runs/35237007970) before the vocabulary follow-up. All test sources compiled; the focused `AsrModelsLocalTests|AsrModelsTests` run had 32 passes, one opt-in real-model test skipped, and no failures. The real-weight transcription checks above ran locally.

The vocabulary follow-up at `0cbec7d` accepts the official v2 vocabulary's extra entries while still requiring every token below the blank ID. Its [full default package build and focused tests](https://github.com/Nathan-Roll1/FluidAudio/actions/runs/35384722308) passed on Xcode: 34 passed, one optional real-model test skipped, zero failures. The added regressions use the published 1,031-entry v2 vocabulary and a temporary copy with one required token removed. Formatting and whitespace checks pass. The earlier real-weight checks above were not rerun for this vocabulary-validation change.


Alex-Wengg:
[P2] The vocabulary count check rejects supported v2 bundles —AsrModels.swift:235–238. The official v2 vocabulary contains 1,031 entries, including every required ID below its blank ID of 1,024. Requiring vocabulary.count == version.blankId causes --local-model-dir … --model- version v2 to fail before loading any models. I reproduced this with the compiled CLI and the  published vocabulary. Check required token coverage while permitting extra entries, and add  regression coverage.

Alex-Wengg (approved):
Reviewed the full change and the final vocabulary-validation follow-up; no blocking findings.

Issue #933, opened 2026-09-19
PocketTTS: the 512-position KV cache forces sentences over 50 tokens to be cut; could the conversion use a longer cache?


arvindvenkataramani:
## Summary

The PocketTTS Core ML conversion has a fixed 512-position KV cache, and because of it FluidAudio cuts every sentence longer than 50 tokens into chunks that are synthesized separately. Prosody restarts at each cut, and where a cut falls between words it is heard as a pause in the middle of a phrase. With `‖` marking the cuts, the 69-token sentence in the steps below is spoken as:

> The humidity feels comfortable despite the approaching storm, ‖ and the air carries that distinctive electric charge that precedes rain — a mix of ozone and wet soil that makes everything feel fresh and alive before the first drops hit ‖ the pavement.

This is the normal case, not an edge case: in two passages of plain English prose (below), all seven sentences measure 52 to 70 tokens, so every one of them is cut.

mlx-audio runs the same model with the same voice and speaks these sentences whole; side by side, the difference is obvious. Kyutai's original Python implementation of Pocket-TTS, which FluidAudio and mlx-audio are both ports of, sits in between: it cuts an oversized sentence only at a comma, semicolon or colon, never between words, and its cache grows with the utterance instead of being fixed.

## Steps to reproduce

```
fluidaudiocli tts --backend pocket --voice alba --seed 42 -o pavement.wav --text "The humidity feels comfortable despite the approaching storm, and the air carries that distinctive electric charge that precedes rain — a mix of ozone and wet soil that makes everything feel fresh and alive before the first drops hit the pavement."
```

Listen for the pause before "the pavement".

## Environment

FluidAudio `main` @ `b68f4847`, `FluidInference/pocket-tts-coreml` v2.1 English. macOS 26.6.2, M4 Max. Comparison: mlx-audio serving `mlx-community/pocket-tts`.

## Request

Is 512 a deliberate limit, for latency, memory or ANE placement? If it is not needed, a longer cache (1024, offered as an opt-in export in the way #925 did for Chatterbox Nano) would let sentences be synthesized whole, as mlx-audio does.

## Related

#931 (fix in #932) corrects the token counts this behaviour depends on. #934 (change in #935) makes the cuts less audible while the cache stays at 512. #584 (closed) addressed how the pieces of a cut sentence are normalized.

---

## Notes — by Claude Code

*Written by Claude Code, an AI assistant working with the reporter. Measured on `b68f4847`.*

[...]

Issue #584, opened 2026-05-07
PocketTTS text chunker produces audible artifacts on long French text (smart apostrophe + clause-cascade)


logicbridgedev:
## Summary

The PocketTTS text chunker in `Sources/FluidAudio/TTS/PocketTTS/Pipeline/PocketTtsSynthesizer.swift` produces two distinct kinds of audible artifacts on French text. Both are reproducible in the current `0.14.2` release.

## Issue 1 — Smart apostrophe (U+2019) treated as sentence boundary

**Symptom**: Smart apostrophes split phrases mid-clause, dropping characters and producing audible glitches at the splits.

**Repro** (debug build, `language=french24L`, voice `fantine`):

Identical text, only the apostrophe character differs:

| Apostrophe | Text | Chunks |
|---|---|---|
| `'` (U+0027 ASCII) | `…une proposition susceptible d'aboutir à une trêve, lancée à la suite…` | **2 chunks** — splits at the comma after `trêve` ✅ |
| `'` (U+2019 smart) | `…une proposition susceptible d'aboutir à une trêve, lancée à la suite…` | **3 chunks** — splits BEFORE `d'aboutir`, mid-phrase ❌ |

Bench-tool log with U+2019:

```
Stream chunk 1/3: 'Sa déclaration intervient après des propos récents de Téhéran évoquant une proposition susceptible.'
Stream chunk 2/3: '        D'aboutir à une trêve.'
Stream chunk 3/3: 'Lancée à la suite des bombardements américains et israéliens du 28 février.'
```

The split before `d'aboutir` chops the phonological unit and capitalizes "D" as a sentence start, breaking French prosody. The chunker's sentence splitter treats U+2019 as `.!?`-equivalent.

**Impact**: Macros / iOS / macOS auto-correct converts ASCII `'` → `'` (U+2019) by default, so essentially every consumer typing French (or English with contractions) hits this. Many imported sources (web, Word, Pages) also use U+2019.

**Suggested fix**: Treat U+2019 as ASCII apostrophe in the sentence splitter; only `.`, `!`, `?` should be sentence-terminating glyphs.

## Issue 2 — Long sentences cascade into mid-clause chunks with broken prosody

[...]

PR #935, opened 2026-09-19
feat(tts/pocket): cut oversized sentences at more kinds of punctuation and balance word splits

### LINKED ISSUE #934: PocketTTS: a sentence that must be cut is cut between words although a dash is available
## Summary

When FluidAudio has to cut a sentence to fit the 50-token chunk limit, it often cuts between two words although better places are available, and a cut between words is heard as a pause in the middle of a phrase. In listening tests with a cloned voice, a cut at a comma could not be heard, because a speaker pauses there anyway, and a cut between words always could.

With `‖` marking the cuts, the sentence in the steps below is spoken as:

> The humidity feels comfortable despite the approaching storm, ‖ and the air carries that distinctive electric charge that precedes rain — a mix of ozone and wet soil that makes everything feel fresh and alive before the first drops hit ‖ the pavement.

The second cut strands "the pavement." as a chunk of its own, although the sentence has a dash a few words earlier, where a speaker would pause.

## Steps to reproduce

```
fluidaudiocli tts --backend pocket --voice alba --seed 42 -o pavement.wav --text "The humidity feels comfortable despite the approaching storm, and the air carries that distinctive electric charge that precedes rain — a mix of ozone and wet soil that makes everything feel fresh and alive before the first drops hit the pavement."
```

Listen for the pause before "the pavement".

## Environment

FluidAudio `main` @ `b68f4847`, `FluidInference/pocket-tts-coreml` v2.1 English. macOS 26.6.2, M4 Max.

## Why this matters

Sentences over 50 tokens are ordinary, so these cuts occur in most paragraphs. This is a small improvement that needs no model change, independent of the larger-cache request in #933.

## Related

#931 (fix in #932) corrects the token counts that cut placement depends on. #584 (closed) changed how the pieces of a cut sentence are normalized and added a fix for a one-word tail.

---

## Notes — by Claude Code

*Written by Claude Code, an AI assistant working with the reporter. Checked against the chunker's output on `b68f4847`.*

- **Why the sentence above is cut where it is.** `splitAtClauseBoundaries` recognises only `, ; :`. The part after the comma is 53 tokens and contains no further comma, so it goes to `splitAtWordBoundaries`, which fills 48 tokens and leaves "the pavement." as a 5-token tail. The fix from #584 moves one word back when the tail is a single word; a two-word tail is not covered.
- **A change is in #935**, with unit tests. Dashes, ellipses and the edges of a bracketed aside also count as clause boundaries, with hyphenated words and numeric ranges kept whole, and word-boundary splits are balanced, so 53 tokens become 26 and 27 instead of 48 and 5. The sentence above is then cut once, at the dash. #935 shows this and two more sentences before and after.
- **Runs of sentence punctuation are pulled apart.** `splitSentences` ends a sentence at every `.`, `!` or `?`, so the input `We waited for the results to come back from the lab... and then the whole building lost power, which nobody had planned for at all. Did it matter?! It did, because every sample in the freezer had to be moved across town before morning.` reaches the model with `the lab. . . and then` and `Did it matter? !`. This happens only when the whole text is over 50 tokens, since shorter text is not split at all. No difference was heard in a listening test. #935 keeps such runs together so the text is passed through as written.
- **The first piece of a cut sentence is given a full stop.** For the input `Clouds are gathering on the western ridge by late afternoon, their gray undersides promising scattered showers that could pass through quickly or linger longer depending on how the wind shifts overnight.`, which is cut at its comma, the model receives `Clouds are gathering on the western ridge by late afternoon.` as the first chunk: the comma is replaced by a full stop. #584 stopped this for the later pieces of a cut sentence, not for the first. Keeping the comma was a marginal improvement in a listening test, and it is not part of #935.

arvindvenkataramani:

Fixes #934. **Stacked on #932**: this branch contains that PR's commit, and only the second commit is new here. Cut placement depends on the correct token counts #932 provides, and the tests reuse its synthetic-vocabulary helper.

PocketTTS synthesizes text in chunks of at most 50 tokens, each spoken separately, so prosody restarts wherever a sentence is cut. In listening tests with a cloned voice, a cut at a comma could not be heard, because a speaker pauses there anyway; a cut between words was always heard, as a pause in the middle of a phrase. The chunker cuts between words more often than it needs to, for two reasons. In the examples `‖` marks a cut, and "before" is `main` with #932 applied.

**1. Only `, ; :` count as places to cut.** A dash or a bracketed aside is passed over.

A 69-token sentence with a dash:

> **Before:** The humidity feels comfortable despite the approaching storm, ‖ and the air carries that distinctive electric charge that precedes rain — a mix of ozone and wet soil that makes everything feel fresh and alive before the first drops hit ‖ the pavement.
>
> **After:** The humidity feels comfortable despite the approaching storm, and the air carries that distinctive electric charge that precedes rain — ‖ a mix of ozone and wet soil that makes everything feel fresh and alive before the first drops hit the pavement.

A 69-token sentence with a bracketed aside and no comma:

> **Before:** The control plane on the Pi (which also serves the dashboard) polls the shard every fifteen minutes and merges whatever services it finds into a ‖ single response that the rest of the platform treats as the source of truth.
>
> **After:** The control plane on the Pi (which also serves the dashboard) ‖ polls the shard every fifteen minutes and merges whatever services it finds into a single response that the rest of the platform treats as the source of truth.

**2. With no punctuation to use, each part is filled to the limit, which strands a short tail.** A two- or three-word chunk gets a prosodic start of its own and is the most noticeable cut of all.

A 52-token sentence with no punctuation:

> **Before:** Nobody should ever need to restart the whole machine just because one small demand loaded model ran out of memory while another bigger one was still busy answering a long question from the ‖ voice loop.
>
> **After:** Nobody should ever need to restart the whole machine just because one small demand loaded model ‖ ran out of memory while another bigger one was still busy answering a long question from the voice loop.

The parts go from 49 and 3 tokens to 26 and 26. This cut is still between words, since the sentence offers nothing better, but it falls at the middle.

### What changed

- **`splitAtClauseBoundaries`**: alongside `, ; :` it now splits at an em dash; at an en dash or hyphen with a space on each side, so hyphenated words and ranges like `10–12` stay whole; after an ellipsis or closing bracket when a space follows, so `),` stays attached; and before an opening bracket.
- **`splitAtWordBoundaries`**: text needing `n` parts aims each at a `1/n` share of its tokens, closing a part at the word boundary nearest that share and never exceeding `maxTokens`. The single-word orphan fix from #584 is kept.
- **`splitSentences`**: a run of sentence punctuation ends one sentence once. Before, `the lab... and then` reached the model as `the lab. . . and then`, and `Did it matter?!` as `Did it matter? !`. No audible difference was found; the change passes the text through as written, as Kyutai's original Python implementation and mlx-audio both do.
- **`normalizeText`**: a trailing dash is stripped from a sentence-start chunk, as trailing `, ; :` already are.

### How this relates to the original

Kyutai's original Python implementation of Pocket-TTS cuts an oversized sentence only at `, ; :` and never between words; a piece still over the limit is left whole, with a logged warning that words may be skipped. FluidAudio cannot leave such a piece whole, because its KV cache is fixed at 512 positions (see #933), so it keeps a word-boundary fallback. This PR makes that fallback rarer, by recognising more punctuation, and less audible, by balancing the parts.

### Verification

- The three "after" results above are the chunker's actual output with the v2.1 English `tokenizer.model`.
- `PocketTtsCutPlacementTests` covers each boundary rule, its exceptions, and the balanced split, with a vocabulary built in the test so no model download is needed. The PocketTTS and tokenizer suites pass, and `swift format lint` is clean.
- By listening, with a cloned voice: the "before" pause ahead of "the pavement" is clearly audible, and cuts at commas are not. The new cuts at a dash and at a bracket rest on the same reasoning and were not judged separately.

*Implemented by Claude Code, an AI assistant, working with the reporter, who verified the results by listening.*


Alex-Wengg:
──────────────────────────────────────────────────────────────────────────────────────────────────

• Found three issues in PR #935, reviewed at ba3c873.

  - P2 — The balancing algorithm can create an unnecessary extra chunk and short tail.
    /tmp/fluidaudio-review-935.JeGMMK/source/Sources/FluidAudio/TTS/PocketTTS/Pipeline/
    PocketTtsSynthesizer.swift:1116 uses a fixed target for every part. Rounding errors accumulate
    instead of adjusting for the remaining tokens. With the real English tokenizer, a 126-token
    input was split into 38, 37, 41, 10, although 38, 48, 40 is feasible. This creates four
    synthesis passes instead of three and recreates the short-tail artifact the PR intends to
    remove. Recalculate the target from the remaining tokens and parts after each cut, or select
    boundaries globally.

  - P2 — Parentheses inside expressions are mistaken for parenthetical asides.
    /tmp/fluidaudio-review-935.JeGMMK/source/Sources/FluidAudio/TTS/PocketTTS/Pipeline/
    PocketTtsSynthesizer.swift:1051 flushes before every ( or [, regardless of surrounding
    whitespace. In a long sentence using the real tokenizer, f(x) became f (x) and produced a
    separate 12-token first chunk. The same affects item(s), Array<Int>(...), and similar text.
    Restrict this boundary to actual spaced asides or preserve the original separator when
    regrouping parts.

  - P2 — maxTokensPerChunk: 0 now crashes with division by zero.
    /tmp/fluidaudio-review-935.JeGMMK/source/Sources/FluidAudio/TTS/PocketTTS/Pipeline/
    PocketTtsSynthesizer.swift:1103 divides by maxTokens. This value comes from a public API
    without validation or a documented positive-only precondition. A multiword input with
    maxTokensPerChunk: 0 terminates the process with Fatal error: Division by zero. Validate it at
    the public entry point or guard here.

  The PR builds and formatting passes. Focused swift test compilation rem

